PR #68, opened 2024-01-08
Add HD wallets support

survived:
Adds HD-wallets support (that requires `hd-wallets` feature) based on `slip_10` crate.

Spec was updated to reflect changes in the protocol.

I also updated keygens to use `utils::blame` which appeared after keygen was written

Diff is huge bc I regenerated precomputed shares


maurges:
> Compared to base branch
>> Diff
>> No changes

I think our bot is broken

cggmp21/src/key_share.rs
```diff
@@ -210,6 +217,44 @@ impl<E: Curve> DirtyIncompleteKeyShare<E> {
         }
         Ok(())
     }
+
+    /// Checks whether HD wallet support was enabled for this key
+    ///
+    /// In order to generate an HD wallet, [`.hw_wallet(true)`](crate::keygen::GenericKeygenBuilder::hd_wallet)
```
maurges: Typo in hw_wallet

cggmp21/src/signing.rs
```diff
@@ -365,6 +406,7 @@ async fn signing_t_out_of_n<M, E, L, D, R>(
     S: &[PartyIndex],
     message_to_sign: Option<DataToSign<E>>,
     enforce_reliable_broadcast: bool,
+    #[cfg(feature = "hd-wallets")] additive_shift: Option<Scalar<E>>,
```
maurges: I think changing the number of parameters depending on enabled features is a terrible idea. Why not make this a mandatory parameter and pass None?

maurges: This would also simplify the code below where we have to ifdef again to get the value, where we could unconditionally do `unwrap_or`

spec/main.tex
```diff
@@ -896,8 +896,13 @@ \subsubsection{Non-Threshold (i.e., $n$-out-of-$n$) Key Generation}
           and set $X_i = x_i \cdot G$.
     \item Sample $\rid_i \gets \{0,1\}^\kappa$.
     \item Compute $(A_i; \tau_i) = \commit{sch}()$.
+    \item {\bf (HD-wallets.)} 
+    \begin{itemize}
+      \item If HD-wallets support enabled, sample local chain code contribution $c_i \gets \{0,1\}^{256}$ (32-bytes string)
+      \item Otherwise, set $c_i = ``"$ (to empty string)
+    \end{itemize}
```
maurges: Might be better to say `set c_i = ⊥` or `nil`? Or even omit mentioning it's set to anything in the otherwise case. I'm concerned about the interpretation of hashing and sending of empty string, but I'm not sure which one reads clearer.

spec/main.tex
```diff
@@ -981,24 +995,28 @@ \subsubsection{Threshold (i.e., $t$-out-of-$n$) Key Generation}
...
       \item For each party $j \neq i$:
         \begin{itemize}
         \item Check that $\vec{S}_j$ has length $t$.
-          \item Assert $V_j = H(\sid\| n\| j\| t\| \rid_j\| \vec{S}_j\| A_j\| u_j)$.
+        \item Assert $V_j = H(\sid\| n\| j\| t\| \rid_j\| \vec{S}_j\| A_j\| u_j \| c_j)$.
         \item Define 
-        $F_j(x) = \sum_{k \in [t]} x^k \cdot S_{j,k}$.
-          \item Assert $\sigma_{j,i} \cdot G  = F_j(i + 1)$.
+          $F_j(x) = \sum_{k \in [t]} x^k \cdot S_{j,k}$.
+        \item Assert $\sigma_{j,i} \cdot G  = F_j(i + 1)$.
```
maurges, line 1011: Should be indented one more level inside `itemize`

spec/main.tex
```diff
@@ -1123,28 +1141,41 @@ \subsubsection{Threshold ($t$-out-of-$n$) Presinging}
...
+  context separation string $\sid$,
+  security level $L$, curve $\E$ with generator $G$.
+
+  {\bf HD-wallets inputs:} additive $\text{shift} \in \Fq$ \footnote{Deriving additive shift is up to specific standard of HD-wallets, e.g. see \cite{bip32} or \cite{slip10}} ($\text{shift} = 0$ disables HD derivation)
 
     \item[\bf Setup.] The key share $K_{S(i)}$ contains $\threshold$, number of key holders~$n$, secret share $x'_{S(i)}$, parties' public shares $\vec X' = (X'_j)_{j \in [n]}$, a map $I : [n] \to \Fq \setminus \{0\}$, Paillier secret key $\sk_{S(i)}$, parties' Paillier keys $\vec N' = (N'_j)_{j \in [n]}$, and parties' auxiliary information $\vec R' = (s_j, t_j, \hat N)_{j \in [n]}$.
 
     \item[\bf Step 1.] Set $\sk_i = \sk_{S(i)}$ and 
-            $\vec R = (R'_{S(j)})_{j \in [n]}$. Then:
+                       $\vec R = (R'_{S(j)})_{j \in [n]}$. Then:
```
maurges, line 1158: Tab character or specific over-indentation?

spec/main.tex
```diff
@@ -1123,28 +1141,41 @@ \subsubsection{Threshold ($t$-out-of-$n$) Presinging}
...
-           \item If shares are additive\footnote{In this case we have $t=n$.} shares of the private key, 
+        \item If shares are additive\footnote{In this case we have $t=n$.} shares of the private key, 
             set $x_{i} = x'_{S(i)}$, $\vec X = (X'_{S(j)})_{j \in [n]}$.
-
         
         \item If shares are Shamir secret shares of the private key: 
-
         \begin{itemize}
-        \item For $j \in [t]$, compute Lagrange coefficient $\lambda_j = \prod\limits_{m \in [t]\setminus\{j\}} \frac{I(S(m))}{I(S(m)) - I(S(j))} \bmod{q}$.
+            \item For $j \in [t]$, compute Lagrange coefficient $\lambda_j = \prod\limits_{m \in [t]\setminus\{j\}} \frac{I(S(m))}{I(S(m)) - I(S(j))} \bmod{q}$.
```
maurges, line 1165: Wow, we really need an autoformatter for latex =) The indentation style is inconsistent in random places

spec/main.tex
```diff
@@ -1123,28 +1141,41 @@ \subsubsection{Threshold ($t$-out-of-$n$) Presinging}
...
             \item Compute $x_{i} = \lambda_i \cdot x'_{S(i)}$.
             \item For $j \in [n]$, compute $X_j= \lambda_j \cdot X'_{S(j)}$; then set $\vec X = \{X_j\}_{j \in [t]}$.
         \end{itemize}
     \end{itemize}
 
-    \item[\bf Step 2.] Call the non-threshold presigning protocol from the previous section using inputs $t, i, x_i, \vec X$, $\sk_i$, $\vec R, \sid, L, \E$.
+    \item[\bf Step 2.] {\bf (HD-wallets)} If HD-wallets support enabled:
+    \begin{itemize}
+      \item Update $X_0 := X_0 + \text{shift} \cdot G$
```
maurges: Might be read that we update the stored `X_0` in the keyshare, when we just use it locally and discard

tests/src/bin/precompute_shares.rs
```diff
@@ -63,20 +63,28 @@ fn precompute_shares_for_curve<E: Curve, R: RngCore + CryptoRng>(
...
-            .collect();
-            let shares = trusted_dealer::builder::<E, SecurityLevel128>(n)
-                .set_threshold(t)
-                .set_pregenerated_primes(primes)
-                .generate_shares(rng)
-                .context("generate shares")?;
-            cache.add_shares(t, n, &shares).context("add shares")?;
+            for hd_enabled in [false, true] {
+                eprintln!(
+                    "t={t:?},n={n},curve={},hd_enabled={hd_enabled}",
```
maurges, line 68: Would be nice to have a test for loading keyshares between different crate features enabled. A bit hard to organize though.

cggmp21/src/signing.rs
survived: Yeah indeed we can let it be None when feature is disabled

survived: But I don't agree that having different number of parameters is "terrible idea". When the function is private and only used in one place, it doesn't matter.

spec/main.tex
survived: The whole document formatting is messed up, there are many places where `itemize` and `item` have the same intent level

spec/main.tex
survived: I leveled it with previous line, like we did in many other places

spec/main.tex
survived: Yep, feel free to suggest!

tests/src/bin/precompute_shares.rs
survived: There are tests for that. On CI, we launch tests with and without hd support, so pregenerated key shares are loaded on both versions of the lib during the tests

cggmp21/src/signing.rs
survived: Updated

spec/main.tex
survived: Changed wording

spec/main.tex
survived: Changed to $\bot$

tests/src/bin/precompute_shares.rs
maurges: I don't see where this is done? Also, I meant having keyshares being generated with hd-wallet feature completely off as well

spec/main.tex
maurges: Yeah, but would be better if we didn't introduce new ones 🙃

spec/main.tex
maurges: Mmmm reads exactly the same to me. Maybe add a note "for this protocol invocation only"?

tests/src/bin/precompute_shares.rs
survived: We do have tests that generate keys, and we do run tests with hd feature disabled, see `test` and `test-hd` in the workflow:

https://github.com/dfns/cggmp21/blob/c003fa41e28a0ff9c3c2919839d7a5663a86f78c/.github/workflows/rust.yml#L28-L46

spec/main.tex
survived: I don't like it. The setup step states that key share contains secret share $x'_{S(i)}$ and list of public shares $\vec X'$. It's different from $x$ and $X$ that we define later. Why would someone assume that changing something derived from the key share must change the key share?

survived: $x$ and $X$ are completely different from the ones stored in the key share, at least because they are multiplied by lagrange coefficient (if it's threshold share). Would the reader assume that key share also needs to be multiplied at lagrnage coef?

spec/main.tex
survived: I did not write this `itemize`, I don't understand what you want from me. I just leveled the misaligned items that had one extra space before them, compared to other items

survived: I'd suggest to move all stylistic suggestions to when we have a formatter

spec/main.tex
maurges: Ah you're right, then it made sense even before. Sorry for bothering

survived: No worries! "Update" does sound a bit misleading, changing it to "set" is probably for the best

tests/src/bin/precompute_shares.rs
maurges: Ah, it's the new workflow, I was looking in the master branch

spec/main.tex
maurges: Sure. I'll resolve all those comments for now

cggmp21/src/signing.rs
```diff
@@ -251,6 +254,8 @@ where
             tracer: None,
             enforce_reliable_broadcast: true,
             _digest: std::marker::PhantomData,
+            #[cfg(feature = "hd-wallets")]
+            additive_shift: None,
```
maurges, line 258: Now I also think we could keep additive shift unconditionally. Builder is an ephemeral struct, so it doesn't seem like a big waste to me. This would simplify the calling of signing functions below.

cggmp21/src/signing.rs
```diff
@@ -425,6 +483,20 @@ where
     };
     debug_assert_eq!(key_share.core.shared_public_key, X.iter().sum::<Point<E>>());
 
+    // Apply additive shift
+    let shift = additive_shift.unwrap_or(Scalar::zero());
+
+    let Shift = Point::generator() * shift;
```
maurges: Super-nitpick: maybe remove this empty line?

tests/src/lib.rs
```diff
@@ -155,3 +164,22 @@ pub fn convert_from_stark_scalar(
     let bytes = x.to_bytes_be();
     generic_ec::Scalar::from_be_bytes(bytes).context("Can't read bytes")
 }
+
+#[cfg(feature = "hd-wallets")]
+pub fn random_derivation_path<E: Curve>(
+    rng: &mut impl rand::RngCore,
+    epub: &cggmp21::slip_10::ExtendedPublicKey<E>,
+) -> (Vec<u32>, generic_ec::Point<E>) {
```
maurges: I would feel more confident in tests if this function didn't return public key, and instead it would be derived close to its usage place. Not sure about this though

tests/src/bin/precompute_shares.rs
maurges: In the initial message I was also thinking of concerns that cargo sometimes still compiles with some features turned on. But let's leave that for later discovery

survived: You can see whether HD feature is on or off by looking at tests. If HD is on, there will be tests with HD suffix in their names

cggmp21/src/signing.rs
survived: I'd prefer to be sure that if HD is off, additive shift is always `None`

survived: If we have field, then we have to make sure it never gets value by analyzing where it's used. If it's never here, it can' be misused

survived: Plus, if we keep the field, the compiler will be complaining that the field is useless (that we have field but never set any value to it)

tests/src/lib.rs
survived: Public key is only used for signature verification in the end of the tests

survived: Well, actually, maybe it's not a bad idea, let me try

survived: See d29536d

cggmp21/src/signing.rs
survived:
```suggestion
    let shift = additive_shift.unwrap_or(Scalar::zero());
    let Shift = Point::generator() * shift;
```

cggmp21/src/signing.rs
maurges: It won't be unused, its value would be passed to functions that expect shift. I agree with your concern about how it should always be None. My concern is that I hate how we have a lot of `ifdef`s now in our code, like in the good(bad) old C.

Feel free to resolve if you feel this is not worth it.

survived: > It won't be unused, its value would be passed to functions that expect shift

I think compiler will complain if you have field, and you read it, but you never write to it (except for initialization)

survived: I feel like `ifdef` usage here is justified

maurges: It is justified, but it doesn't make it less unseemly

maurges (approved):

PR #69, opened 2024-01-08
Fix docs build issue occurred on docs.rs

survived:
Issue was that docs.rs couldn't locate `katex-header.html` in the crate. Creating a symlink should fix that.


maurges (approved):
A lot of changes unrelated to what you declare, but ok

survived:
I also added badges and addressed clippy errors

PR #72, opened 2024-01-30
Restructure library

survived:
CGGMP21 DKG has been proved to be UC-secure, and thus it can be safely reused with other UC-secure TSS implementations. To make it reusable, this PR moves DKG implementation into separate crate `cggmp21-keygen`, and key share definition into separate crate `key-share`


Cargo.toml
```diff
@@ -2,5 +2,11 @@
...
 members = [
     "cggmp21",
+    "cggmp21-keygen",
+    "key-share",
     "tests",
 ]
+
+[patch.crates-io.generic-ec]
+git = "https://github.com/dfns/generic-ec"
+branch = "small-impos"
```
survived: Need to merge that first

cggmp21-keygen/src/rng.rs
survived, line 1: We really should move it into another crate

tests/tests/it/main.rs
survived, line 1: This optimization only saved 50s: tests went from 17m to 16m19s

maurges: You mean collapsing integration tests into one build unit? It's mainly done to improve _build_ time performance I believe. That's the reason I did it in signers anyway

survived: The article you sent claimed it could significantly improve tests time as well. It did but not significantly

survived: Well, actually, 50s is something, it's almost a minute! I just shared an observation

cggmp21-keygen/src/lib.rs
```diff
@@ -0,0 +1,318 @@
...
+#[doc(no_inline)]
+pub use self::msg::{non_threshold::Msg as NonThresholdMsg, threshold::Msg as ThresholdMsg};
+
+/// Defines default choice for digest and security level used across the crate
+mod default_choice {
+    pub type Digest = sha2::Sha256;
+    pub type SecurityLevel = crate::security_level::SecurityLevel128;
+}
+
+#[doc = include_str!("../../cggmp21/docs/mpc_message.md")]
```
maurges: Maybe move docs to repo root?

cggmp21-keygen/src/non_threshold.rs
maurges, line 1: Oh my god, why can't git detect file rename here

cggmp21-keygen/src/progress.rs
maurges, line 1: Symlink this one similarly to hex_or_bin?

maurges: Oh wait, it's gone from main protocol? Weren't we also using it to profile there?

maurges:
Did you forget to remove keygen files from `cggmp21/`? It seems they are no longer part of the module tree

key-share/src/lib.rs
```diff
@@ -0,0 +1,323 @@
...
+/// * Signer index `i` is less than `n`
+/// * Signer public commitment matches the secret share
+/// * Threshold value is within range `2 <= t <= n`
+/// * All signers commitments sum up to public key
+///
+/// It's impossible to obtain [`CoreKeyShare`] for the key share that doesn't meet above requirements.
+///
+/// Only immutable access to the key share is provided. If you need to change content of the key share,
+/// you need to obtain dirty key share via [`Valid::into_inner`], modify the key share, and validate it
+/// again to obtain `CoreKeyShare`.
```
maurges, line 41: I don't particularly care either way, but why did you decide to pull it from doc file to inline doc?

key-share/src/lib.rs
```diff
@@ -0,0 +1,323 @@
...
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+#[cfg_attr(feature = "serde", serde(bound = ""))]
+pub struct DirtyCoreKeyShare<E: Curve> {
+    /// Guard that ensures curve consistency for deseraization
+    pub curve: CurveName<E>,
+    /// Index of local party in key generation protocol
+    pub i: u16,
+    /// Public key corresponding to shared secret key. Corresponds to _X_ in paper.
+    #[cfg_attr(feature = "serde", serde(with = "As::<generic_ec::serde::Compact>"))]
+    // #[cfg_attr(feature = "serde", serde_as(as = "generic_ec::serde::Compact"))]
```
maurges, line 83: ^

key-share/src/lib.rs
```diff
@@ -0,0 +1,323 @@
...
+    pub shared_public_key: Point<E>,
```
maurges: Middle line

key-share/src/valid.rs
```diff
@@ -0,0 +1,242 @@
...
+/// }
+/// ```
+pub trait ValidateFromParts<Parts>: Validate {
+    /// Validates parts
+    ///
+    /// Note: implementation **must** guarantee that if `T::validate_parts(parts).is_ok()` then `T::from_parts(parts).is_valid().is_ok()`
+    fn validate_parts(parts: &Parts) -> Result<(), Self::Error>;
+    /// Constructs `Self` from parts
+    fn from_parts(parts: Parts) -> Self;
+}
```
maurges, line 154: Do we really need this trait if there's going to be only one implementor? I'd rather have a trait `FromParts` implemented for valid and dirty shares, one checking for validity and the other not. This starts to look suspiciously similar to what we had before with TryFrom... But I'm really _not_ suggesting to go back to that, I like when my go-to-definition actually works

maurges: Ah, I forgot that AuxInfo is not part of keyshare anymore and can be anything, which makes a `from_parts` a nice convenient method on Valid. Although I don't care about it for KU23 for example, because core share is all I need.

cggmp21-keygen/src/non_threshold.rs
survived: Yeah it can't do that somethimes. I didn't change anything in this file or `threshold.rs`

cggmp21-keygen/src/progress.rs
survived: We expose this module (we have to if we want someone to do profiling), so cggmp21::progress just re-exports `cggmp21_keygen::progress`.

survived: It probably should be a dedicated crate actually, maybe we'll move it soon

key-share/src/valid.rs
survived: If you don't care, don't use it :D For cggmp21 it's needed

cggmp21-keygen/src/non_threshold.rs
maurges: I had to look in all files in local vim. Which was actually a nicer experience than using github.

But here it doesn't detect change because you forgot to remove the original file!

survived: Ah indeed! I'll remove

key-share/src/lib.rs
survived: Originally, I was using doc files to avoid repeating the same stuff in two different places. In this PR, I changed the docs so there's no repetition between `DirtyCoreKeyShare` and `CoreKeyShare`

cggmp21-keygen/src/non_threshold.rs
survived: Now it sees that the file was moved

maurges (approved):

PR #75, opened 2024-02-08
HD wallets: update the spec

survived:
I forgot to update the spec in #74. Also adding `shell.nix` that helps compiling latex, and a minor change in `Presignature::set_derivation_path` method.

### DIFF
--- a/shell.nix
+++ b/shell.nix
@@ -1,0 +1,23 @@
+let # Rust
+    rust_overlay = import (builtins.fetchTarball "https://github.com/oxalica/rust-overlay/archive/master.tar.gz");
+    pkgs = import <nixpkgs> { overlays = [ rust_overlay ]; };
+    rustVersion = "1.75.0";
+    rust = pkgs.rust-bin.stable.${rustVersion}.default.override {
+      extensions = [
+        "rust-src" # for rust-analyzer
+      ];
+    };
+    # Latex
+    tex = (pkgs.texlive.combine {
+      inherit (pkgs.texlive) scheme-small
+        collection-mathscience preprint;
+    });
+    
+
+in pkgs.stdenv.mkDerivation {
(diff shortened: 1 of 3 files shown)


shell.nix
```diff
@@ -0,0 +1,23 @@
+let # Rust
+    rust_overlay = import (builtins.fetchTarball "https://github.com/oxalica/rust-overlay/archive/master.tar.gz");
+    pkgs = import <nixpkgs> { overlays = [ rust_overlay ]; };
+    rustVersion = "1.75.0";
+    rust = pkgs.rust-bin.stable.${rustVersion}.default.override {
+      extensions = [
+        "rust-src" # for rust-analyzer
+      ];
+    };
```
maurges, line 9: If you want to use newer rustc, you could instead import nixpkgs-unstable from some date (like today). Not sure which way is better. I usually use the method with unstable.

maurges: Example:
```
import (builtins.fetchTarball {
  # Descriptive name to make the store path easier to identify
  name = "nixos-unstable-2018-09-12";
  # Commit hash for nixos-unstable as of 2018-09-12
  url = "https://github.com/nixos/nixpkgs/archive/ca2ba44cab47767c8127d1c8633e2b581644eb8f.tar.gz";
  # Hash obtained using `nix-prefetch-url --unpack <url>`
  sha256 = "1jg7g6cfpw8qvma0y19kwyp549k1qyf11a5sg6hvn6awvmkny47v";
}) {}
```

survived: I have no preference as long as it works. If you can change it and the project keeps compiling, go ahead!

maurges (approved):

shell.nix
maurges: Me neither, so let's leave as is

PR #79, opened 2024-02-19
Add trusted dealer to key-share itself

### LINKED ISSUE #78: Suggestion: key-share should export generate_core_share()
Ran into it while writing test for the new protocol. Of course we'll need to add an `spof` feature for that as well.

maurges:
Closes #78

Trusted dealer in cggmp now uses that for some calls.

Of note: when generating shares, we generated useless rid. I removed that.


cggmp21/src/trusted_dealer.rs
```diff
@@ -51,16 +49,15 @@ pub fn builder<E: Curve, L: SecurityLevel>(n: u16) -> TrustedDealerBuilder<E, L>
     TrustedDealerBuilder::new(n)
 }
 
+type Inner<E> = key_share::trusted_dealer::TrustedDealerBuilder<E>;
```
survived: A bit confusing. Maybe `CoreBuilder` or `CoreDealer`?

key-share/src/trusted_dealer.rs
```diff
@@ -0,0 +1,182 @@
...
+//! dealer is mainly intended to be used in tests.
+//!
+//! ## Example
+//! Import a key into 3-out-of-5 TSS:
+//! ```rust,no_run
+//! # use rand_core::OsRng;
+//! # let mut rng = OsRng;
+//! use generic_ec::{curves::Secp256k1, SecretScalar};
+//!
+//! let secret_key_to_be_imported = SecretScalar::<Secp256k1>::random(&mut OsRng);
```
survived: Did you mean `&mut rng`?  `let mut rng = OsRng` is defined above

key-share/src/trusted_dealer.rs
```diff
@@ -0,0 +1,182 @@
...
+                .collect::<Vec<_>>();
+            (pk, shares)
+        } else {
+            let mut shares = std::iter::repeat_with(|| SecretScalar::<E>::random(rng))
+                .take((self.n - 1).into())
+                .collect::<Vec<_>>();
+            shares.push(SecretScalar::new(
+                &mut (shared_secret_key - shares.iter().sum::<Scalar<E>>()),
+            ));
+            let pk = shares.iter().map(|x_j| Point::generator() * x_j).sum();
```
survived: I've just realized that it'll be more efficient to sum all the shares, and only then to multiply by G

survived:
Could you also add `reconstruct_secret_key` to `key_share`? I guess, it would accept `&[impl AsRef<CoreKeyShare<E>>]` instead of `&[impl AnyKeyShare<E>]`

https://github.com/dfns/cggmp21/blob/09d96ab269fa7b2f3bcb311393b0ae47e8786f76/cggmp21/src/key_share.rs#L370-L372

key-share/src/trusted_dealer.rs
maurges: I just copy-pased the code from cggmp =) I'll fix it

cggmp21/src/key_share.rs
```diff
@@ -336,31 +332,16 @@ pub trait AnyKeyShare<E: Curve>: sealed::Sealed {
...
     }
 }
 
 impl<E: Curve, L: SecurityLevel> sealed::Sealed for KeyShare<E, L> {}
-impl<E: Curve, L: SecurityLevel> AnyKeyShare<E> for KeyShare<E, L> {
-    fn core(&self) -> &DirtyIncompleteKeyShare<E> {
-        &self.core
-    }
-}
+impl<E: Curve, L: SecurityLevel> AnyKeyShare<E> for KeyShare<E, L> {}
```
survived: Can you do a blanket implementation?

```rust
impl<E: Curve, T> AnyKeyShare for T
where
    T: AsRef<IncompleteKeyShare<E>>
{}
```

we won't need `sealed::Sealed` then

survived: Still not clear if we need this trait at all. If you think it improves readability, we can keep it, but rather than this I don't see how it's useful

maurges: Good idea, done

survived (dismissed):

key-share/src/lib.rs
```diff
@@ -419,3 +426,94 @@ impl<T> From<ValidateError<T, InvalidCoreShare>> for InvalidCoreShare {
...
+/// generation). Accepts both [`KeyShare`] and [`IncompleteKeyShare`].
+/// Returns error if input is invalid.
+///
+/// Note that, normally, secret key is not supposed to be reconstructed, and key
+/// shares should never be at one place. This basically defeats purpose of MPC and
+/// creates single point of failure/trust.
+#[cfg(feature = "spof")]
+pub fn reconstruct_secret_key<E: Curve>(
+    key_shares: &[impl AsRef<CoreKeyShare<E>>],
+) -> Result<SecretScalar<E>, ReconstructError> {
```
survived: Doc is outdated. There's no "`KeyShare` and `IncompleteKeyShare`" in the `key_share` crate

survived: Can you add a CI job so it catches that stuff? This is an [example](https://github.com/dfns/generic-ec/blob/ba27154f93ec2502f71063056e1da35039140a69/.github/workflows/rust.yml#L81-L90) from generic-ec:

```
check-doc:
  runs-on: ubuntu-latest
  steps:
  - uses: dtolnay/rust-toolchain@nightly
  - uses: actions/checkout@v3
  - uses: Swatinem/rust-cache@v2
    with:
      cache-on-failure: "true"
  - name: Check docs
    run: RUSTDOCFLAGS="--cfg docsrs -D warnings" cargo +nightly doc --workspace --all-features --no-deps --exclude nostd-example
```

maurges: Added CI, it required fixing some unrelated docs as well

survived (approved):

PR #80, opened 2024-02-20
Add `KeyInfo`, redesign `Valid<T>`

survived:
* `KeyInfo` contains public information about the key. It's useful in FROST.
* `Valid<T>` now transitively trusts that if `T` contains `K`, then `K` must have been validated when `T` was validated. That allows us to have `as_ref :: &Valid<T> -> &Valid<K>`

### DIFF
--- a/key-share/src/lib.rs
+++ b/key-share/src/lib.rs
@@ -40,6 +40,20 @@ pub use self::valid::{Valid, Validate, ValidateError, ValidateFromParts};
 /// you need to obtain dirty key share via [`Valid::into_inner`], modify the key share, and validate it
 /// again to obtain `CoreKeyShare`.
 pub type CoreKeyShare<E> = Valid<DirtyCoreKeyShare<E>>;
+/// Public Key Info
+///
+/// Type alias to [`DirtyKeyInfo`] wrapped into [`Valid<T>`](Valid), meaning that the key info
+/// has been validated that:
+/// * Number of signers `n` doesn't overflow [`u16::MAX`], and that n >= 2
+/// * Threshold value is within range `2 <= t <= n`
+/// * All signers commitments sum up to public key
+///
+/// It's impossible to obtain [`KeyInfo`] that doesn't meet above requirements.
+///
+/// Only immutable access to the key info is provided. If you need to change content of the key info,
+/// you need to obtain dirty key info via [`Valid::into_inner`], modify the key info, and validate it
+/// again to obtain [`KeyInfo`].
+pub type KeyInfo<E> = Valid<DirtyKeyInfo<E>>;
 
 #[cfg(feature = "serde")]
 use serde_with::As;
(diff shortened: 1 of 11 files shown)


key-share/src/lib.rs
```diff
@@ -119,45 +133,90 @@ pub struct VssSetup<E: Curve> {
...
 }
 
+/// Public Key Info
+///
+/// Contains public information about the TSS key, including shared public key, commitments to
+/// secret shares and etc.
+#[derive(Clone, Debug)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+#[cfg_attr(feature = "serde", serde(bound = ""))]
+pub struct DirtyKeyInfo<E: Curve> {
```
maurges: Do we use udigest in this crate yet? Would be convenient to have

survived: Yep good idea! Made it digestable

key-share/src/valid.rs
```diff
@@ -46,13 +46,27 @@ where
...
 impl<T> Valid<T> {
     /// Returns wraped validated value
     pub fn into_inner(self) -> T {
         self.0
     }
+
+    /// Returns a wrapped reference to validated data
+    pub fn as_ref(&self) -> Valid<&T> {
+        Valid(&self.0)
+    }
```
maurges: Conflicts with AsRef::as_ref which is automatically imported. Maybe use similar name to [`Option::as_deref`](https://doc.rust-lang.org/std/option/enum.Option.html#method.as_deref)?

key-share/src/valid.rs
```diff
@@ -46,13 +46,27 @@ where
...
     }
+
+    /// Constructs `Self` from its [`ValidProjection`]
+    pub fn from<K>(value: Valid<K>) -> Self
+    where
+        T: ValidProjection<K>,
+        K: Validate,
+    {
+        Self(value.0.into())
+    }
```
maurges: Again, I think it's better to avoid naming conflicts with `From::from` since we use those instances too. How about `from_projection`?

key-share/src/valid.rs
```diff
@@ -153,6 +174,9 @@ pub trait ValidateFromParts<Parts>: Validate {
     fn from_parts(parts: Parts) -> Self;
 }
 
+/// Marker trait stating that `Self` is projection of `T`: if `T` is valid, then `Self::from(T)` is valid as well
+pub trait ValidProjection<T: Validate>: Validate + From<T> {}
+
```
maurges: It seems we want to have instances for cggmp keyshare+aux too?

key-share/src/valid.rs
survived: It already has similar name to [`Option::as_ref`](https://doc.rust-lang.org/std/option/enum.Option.html#method.as_ref)...

survived: `as_deref` implies different sematics

key-share/src/valid.rs
survived: Well, we don't have any `From<T>` implementations on `Valid`, so it should not lead to conflicts

survived: I was considering `from_proj` name, but it's uglier

key-share/src/valid.rs
survived: What do you mean exactly?

key-share/src/valid.rs
survived: It does conflict with `as_ref`, but the `AsRef` trait was implemented primarily to be used when some function accepts `impl AsRef<KeyShare>`. That still works.

key-share/src/valid.rs
survived: Ah probably you mean that we should add
```rust
impl ValidProjection<KeyShare> for IncompleteKeyShare {}
impl ValidProjection<KeyShare> for AuxData {}
```

that does make sense

survived: Tho now I better see a pattern, probably what we want is ValidationLens? Wdyt?

```rust
/// If `Self` contains `T` and `Self` was validated, then `T` is also valid
pub trait ValidationLens<T> {
    fn lens(&self) -> &T;
}
```

key-share/src/valid.rs
maurges: Huh, you're right, I'm reading the docs жопой again. Well, since it works for Option, sure

key-share/src/valid.rs
maurges: We don't? Right, all the From instance are for dirty data. I still slightly prefer a more descriptive name, but I'll leave the choice to you.

key-share/src/valid.rs
survived: and then 

```rust
impl<T: Validate> Valid<T> {
    pub fn lens<L>(&self) -> Valid<&L>
    where
        T: ValidationLens<L>
    { /* ... */ }
}

Comment (survived) [reply key-share/src/valid.rs]: naming is confusing tho

Comment (maurges) [reply key-share/src/valid.rs]: BTW, I suggest renaming it with a preposition to show who projects to whom:  
`pub trait ValidProjectionFrom<T: Validate>: Validate + From<T>`  
Thus we get `impl ValidProjectionFrom<KeyShare> for AuxData`

Comment (maurges) [reply key-share/src/valid.rs]: As for lens, I think From is more useful. For example, in key-share we have (or will have) functions that accept `CoreKeyShare ~ Valid<DirtyCoreShare>`, but we want to call them somehow with `KeyShare ~ Valid<DirtyKeyShare>`. Thus we need conversion `KeyShare -> CoreKeyShare`.

Lens instead gives us `&KeyShare -> Valid<&DirtyKeyShare>` which is useless. We could recover some of the usefulness with `cloned :: Valid<&T> -> Valid<T>`, but this just rediscovers From::from again

Comment (maurges) [reply key-share/src/valid.rs]: Though I agree that the function `project :: forall T S. ValidProjectionFrom<T> S => Valid<T> -> Valid<S>` must exist; I just tried to write some conversions without it and failed

File: key-share/src/valid.rs (hunk 3 above)
Comment (survived) [reply key-share/src/valid.rs]: It's been removed as we have AsRef thing now

File: key-share/src/valid.rs (hunk 4 above)
Comment (survived) [reply key-share/src/valid.rs]: I replaced the whole thing with AsRef trick

File: key-share/src/valid.rs [hunk 5]
```diff
@@ -47,13 +122,19 @@ where
...
+    fn from_ref_unchecked(value: &T) -> &Self {
+        #[cfg(debug_assertions)]
+        value
+            .is_valid()
+            .expect("debug assertions: value is invalid, but was assumed to be valid");
+
+        // SAFETY: &T and &Valid<T> have exactly the same in-memory representation
+        // thanks to `repr(transparent)`, so it's sound to transmute the references.
+        // Note also that input and output references have exactly the same lifetime.
+        unsafe { core::mem::transmute(value) }
```
maurges: Maybe for debug assertions check the resulting value for validity as well?

survived: Oh, actually, I meant to check correctness of the output

survived: It probably doesn't make much sense to check before conversion

survived: Wait, but it does check correctness of output value when doing conversion though `as_ref`

survived: This method doesn't coerce the type, it only changes `&T -> &Valid<T>`

survived: `sub_value` has type `&K`, so `&K` will be validated for correctness by `from_ref_unchecked`

https://github.com/dfns/cggmp21/blob/2b2ec834ee55816e5feb7d555340d9d8ba7500e1/key-share/src/valid.rs#L162-L171

key-share/src/lib.rs
```diff
@@ -206,15 +265,47 @@ fn validate_vss_key_share<E: Curve>(
...
+    /// Returns share index of j-th signer
+    ///
+    /// * For additive shares, share id is `j+1`
+    /// * For VSS-shares, share id is scalar $I_j$ such that $x_j = F(I_j)$ where
+    ///   $F(x)$ is polynomial co-shared by the signers
+    ///
+    /// Note: if you have no idea what it is, probably you don't need it.
+    pub fn share_id(&self, j: u16) -> Option<NonZero<Scalar<E>>> {
+        let Some(vss_setup) = self.vss_setup.as_ref() else {
+            return if usize::from(j) < self.public_shares.len() {
```
maurges: You can omit return by using `let Some ... else ...` above

key-share/src/lib.rs
```diff
@@ -206,15 +265,47 @@ fn validate_vss_key_share<E: Curve>(
...
+                #[allow(clippy::expect_used)]
+                Some(
+                    NonZero::from_scalar(Scalar::one() + Scalar::from(j))
+                        .expect("1 + i_u16 is guaranteed to be nonzero"),
+                )
```
maurges: Maybe return None on error case as well? To not pollute the resulting library with strings

survived: I can do `unwrap` instead

survived: I think it's better to panic when happens something what we didn't expect to happen

key-share/src/lib.rs
survived: Changed

key-share/src/lib.rs
maurges: Unwrap still pollutes, so let's just leave as `expect` as we've done everywhere else here

key-share/src/valid.rs
```diff
@@ -5,8 +5,71 @@ use std::fmt;
...
 /// `Valid<T>` provides only immutable access to `T`. For instance, if you want to change content of `T`, you
-/// need to [deconstruct](Valid::into_inner) it, do necessary modifications, and then validate it again using `TryFrom`.
+/// need to [deconstruct](Valid::into_inner) it, do necessary modifications, and then validate it again.
+///
+/// ## Transitive "valideness" through `AsRef`
+/// `Valid<T>` assumes that if `T` implements `AsRef<K>` and `K` can be validated (i.e. `K` implements [`Validate`]),
+/// then `K` has been validated when `T` was validated. Thus, if you have value of type `Valid<T>`, you can obtain
+/// `&Valid<K>` via `AsRef` trait.
+///
+/// Example of transitive valideness is demostrated below:
```
maurges, line 15: I'm not sure such a long example is necessary. for me this feature feels very natural. Instead I would just mention like "An example of transitive validness is obtaining `KeyInfo` from `CoreKeyShare`"

key-share/src/valid.rs
maurges: Right, I just got confused as to what Self and T refer to. Everything is good here.

key-share/src/valid.rs
survived: Well, I think it's because you're deeply familiar with the code and the project already. I feel that it's important to elaborate on that to make it clear how exactly the transitive trust works. Refering to other existing structs would require reader to go to these structs and understand their definition and figure out what we mean by transitive trust.

survived: The example I wrote is simplified all-contained version of these structs, it should be easier to read.

maurges (dismissed):

maurges:
Trying this for my branch: we should now make AnyKeyShare::core give `Valid<CoreShare>` instead of DirtyCoreShare. Though I can do it myself. Another pain point: there's no `AsRef<T> for T`, so I still have to clone

maurges (dismissed):

survived:
> Another pain point: there's no AsRef<T> for T, so I still have to clone

You need to manually implement it for each `T`, it's a bit more code, but it's common practice with AsRef trait

> we should now make AnyKeyShare::core

I actually have feeling that we probably don't need AnyKeyShare anymore, `impl AsRef<CoreKeyShare>` should pretty much replace it.

maurges (approved):

PR #81, opened 2024-02-28
Fix publish

maurges:


.github/workflows/rust.yml
```diff
@@ -92,3 +96,32 @@ jobs:
...
+  # TODO: publish the crates above and add the check below (doesn't work without them)
+  #check-publish-cggmp21:
+  #  runs-on: ubuntu-latest
+  #  steps:
+  #  - uses: actions/checkout@v3
+  #  - uses: Swatinem/rust-cache@v2
+  #    with:
+  #      cache-on-failure: "true"
+  #  - name: Dry-run publish
+  #    run: cargo publish --dry-run -p cggmp21
```
survived: Use a matrix instead: see [example](https://github.com/dfns/generic-ec/blob/df7501ec4d2d686be96fc8f5bc8f75599ceab91a/.github/workflows/rust.yml#L121-L136) in generic-ec

survived:
Why is it draft? Is there something else you wanted to do?

maurges:
Because I was force-pushing changes to test workflow and not leave many commits

survived (approved):

PR #82, opened 2024-02-28
Prohibit `sk=0` and `x_j=0`

survived:
Prohibit the key shares that contain zero public key / secret share / commitment to other signer share

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -7,3 +7,6 @@ members = [
     "tests",
 ]
 
+[patch.crates-io.generic-ec]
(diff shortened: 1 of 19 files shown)


Cargo.toml
```diff
@@ -7,3 +7,6 @@ members = [
     "tests",
 ]
 
+[patch.crates-io.generic-ec]
+git = "https://github.com/dfns/generic-ec"
+branch = "nonzero-cmp"
```
survived: We should merge that first

maurges (approved):

Issue #83, opened 2024-03-04
V0.1.1: Final outgoing signing broadcast does not always finish sending


tbraun96:
To fix this issue, see this diff here: https://github.com/dfns/cggmp21/compare/v0.1.1...webb-tools:cggmp21:v0.1.1-fix

By yielding to the async runtime, we allow the outgoing messages to completely get flushed to the outgoing stream. Without this yield, I have recorded peers sometimes not sending this final broadcast message, despite other peer(s) finishing the protocol.

Note: this was all tested while using futures channels for the `Delivery` implementation.

maurges:
This is a strange problem, especially seeing how another yield happens right after this one in receiving messages. Does it also happen in keygen? Are you sure this is not a quirk of your runtime? Can you try and create a reproduction?

tbraun96:
We are using the tokio runtime. I have not noticed this issue in the keygen.

> This is a strange problem, especially seeing how another yield happens right after this one in the receiving messages

I do not see this yield?


```rust
    outgoings
        .send(Outgoing::broadcast(Msg::Round4(MsgRound4 {
            sigma: partial_sig.sigma,
        })))
        .await
        .map_err(IoError::send_message)?;
    tracer.msg_sent();
    // runtime.yield_now().await; < --- proposed new code

    // Output
    tracer.named_round_begins("Signature reconstruction");

    tracer.receive_msgs();
    let partial_sigs = rounds
        .complete(round4)
        .await
        .map_err(IoError::receive_message)?;
    tracer.msgs_received();
    let sig = {
        let r = NonZero::from_scalar(partial_sig.r);
        let s = NonZero::from_scalar(
            partial_sig.sigma + partial_sigs.iter().map(|m| m.sigma).sum::<Scalar<E>>(),
        );
        Option::zip(r, s).map(|(r, s)| Signature { r, s }.normalize_s())
    };
    let sig_invalid = match &sig {
        Some(sig) => sig.verify(&pk, &message_to_sign).is_err(),
        None => true,
    };
    if sig_invalid {
        // Following the protocol, party should broadcast additional proofs
        // to convince others it didn't cheat. However, since identifiable
        // abort is not implemented yet, this part of the protocol is missing
        return Err(SigningAborted::SignatureInvalid.into());
    }
    let sig = sig.ok_or(SigningAborted::SignatureInvalid)?;

    tracer.protocol_ends();
    Ok(ProtocolOutput::Signature(sig))

```

[...]

PR #84, opened 2024-03-05
Fix outgoing message loss bug with runtime yield

tbraun96:
Solves #83


survived:
Closing per discussion in #83

PR #86, opened 2024-03-12
Update changelog

survived:


cggmp21/CHANGELOG.md
```diff
@@ -1,5 +1,31 @@
...
 
-## Version history lost to time
+## v0.2.0
+* Add support of HD wallets compatible with BIP-32 and SLIP-10 [#68],
+  [#74], [#75]
+* Restructure the library: move reusable structs and functionalities into separate
+  sub-crates [#72], [#76], [#77], [#79]
+* Move public info of the key share into separate struct `KeyInfo` [#80]
+* Prohibit key shares with zero secret share or secret key [#82]
+* Add specs and audit report [#70], [#85]
```
maurges, line 10: Wow, keeping a changelog is hard. From what I see from history, PR 74 and up were not there when we bumped the version to 0.2.0. Why didn't we release with the bump I don't remember anymore. I would prefer to keep the version in changelog and in cargo.toml unbumped until we actually release

maurges: Righhht, I think I just had problems making a release and postponed it until later. This is silly.

maurges (approved):

PR #87, opened 2024-03-12
Bump key-share version and add its changelog

maurges:
Should fix releasing keygen and cggmp21

### DIFF
--- a/key-share/CHANGELOG.md
+++ b/key-share/CHANGELOG.md
@@ -1,5 +1,15 @@
 # Changelog
 
+## v0.2.0
+* Add support of HD wallets compatible with BIP-32 and SLIP-10 [#68],
+  [#74], [#75]
+* Prohibit key shares with zero secret share or secret key [#82]
+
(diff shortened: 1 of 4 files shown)


key-share/CHANGELOG.md
```diff
@@ -1,5 +1,15 @@
 # Changelog
 
+## v0.2.0
+* Add support of HD wallets compatible with BIP-32 and SLIP-10 [#68],
+  [#74], [#75]
+* Prohibit key shares with zero secret share or secret key [#82]
```
survived, line 6: Should we remove the same log messages from cggmp21/changelog ?

maurges: It touched cggmp21 in that we had to alter it a bit to support it, so I think it should be in both

survived: Well, in the future, I don't think we'll be duplicating changelog messages between the logs

survived: we're too lazy for that

survived (approved):

Issue #89, opened 2024-03-15
Bincode: Cannot serialize IncompleteKeyShare<E>


tbraun96:
When running this function, serialization using bincode2 fails:

```rust
pub async fn run_and_serialize_keygen<
    'r,
    E: Curve,
    S: SecurityLevel,
    H: Digest<OutputSize = U32> + Clone + Send + 'static,
    D,
    R,
>(
    tracer: &mut PerfProfiler,
    eid: dfns_cggmp21::ExecutionId<'r>,
    i: u16,
    n: u16,
    t: u16,
    party: MpcParty<Msg<E, S, H>, D>,
    mut rng: R,
) -> Result<Vec<u8>, JobError>
where
    D: Delivery<Msg<E, S, H>>,
    R: RngCore + CryptoRng,
{
    let builder = KeygenBuilder::<E, S, H>::new(eid, i, n);
    let incomplete_key_share = builder
        .set_progress_tracer(tracer)
        .set_threshold(t)
        .start(&mut rng, party)
        .await
        .map_err(|err| JobError {
            reason: format!("Keygen protocol error (run_and_serialize_keygen): {err:?}"),
        })?;
    bincode2::serialize(&incomplete_key_share).map_err(|err| JobError {
        reason: format!("Keygen protocol error (run_and_serialize_keygen - serde): {err:?}"),
    })
}
```

The internal implementation for Serialize does not work.

Using the latest version of main/master branch.

drewstone:
With `bincode` too it persists. I've reproduced it in the repo in one of the tests: https://github.com/dfns/cggmp21/pull/90

survived:
It seems that this happens bacause bincode isn't compatible with `#[serde(flatten)]` tag. For instance, this also fails to serialize:

```rust
#[derive(serde::Serialize, serde::Deserialize)]
struct Foo {
    i: u16,
    #[serde(flatten)]
    bar: Bar,
    x: u32,
}

#[derive(serde::Serialize, serde::Deserialize)]
struct Bar {
    thing: String,
    list: Vec<u8>,
}

#[test]
fn this_fails() {
    let foo = Foo {
        i: 1,
        bar: Bar {
            thing: "123".into(),
            list: vec![1, 2, 3],
        },
        x: 7,
    };

    bincode::serialize(&foo).unwrap();
}
```

[...]

PR #91, opened 2024-03-18
Add note about serialization to the key share docs

survived:
#89 reminded me that we never explicitly mentioned which serialization backeds we support. PR adds this information to the docs.

### DIFF
--- a/key-share/src/lib.rs
+++ b/key-share/src/lib.rs
@@ -97,9 +97,12 @@ use serde_with::As;
 /// It's unlikely, but at some point, we might introduce a breaking change into the serialization format. In this case,
 /// we'll announce it and publish the migration instructions.
 ///
-/// Not every serde backend is supported. We strongly advise using either [`serde_json`](https://docs.rs/serde_json/),
-/// if verbose/human-readable format is needed, or [`ciborium`](https://docs.rs/ciborium/latest/ciborium/), if you'd
-/// like to opt for binary format. Other serialization backends are not tested and disadvised to use.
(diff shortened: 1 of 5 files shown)


key-share/src/lib.rs
```diff
@@ -88,6 +88,21 @@ use serde_with::As;
...
+/// We make our best effort to keep serialization format the same between the versions (even with breaking changes),
+/// and so far we've never introduced breaking change into the serialization format. This ensures that newer versions
+/// of library are able to deserialize the key shares produced by the old version version of the library.
+///
+/// It's unlikely, but at some point, we might introduce a breaking change into the serialization format. In this case,
+/// we'll announce it and publish the migration instructions.
+///
+/// Not every serde backend is supported. We strongly advise using either [`serde_json`](https://docs.rs/serde_json/),
+/// if verbose/human-readable format is needed, or [`ciborium`](https://docs.rs/ciborium/latest/ciborium/), if you'd
+/// like to opt for binary format. Other serialization backends are not tested and disadvised to use.
```
maurges: Maybe add an explaination? "Not every serde backend supports some features we use to ensure backwards compatability, which you will only detect at runtime. We advise you to use either ..., which are tested to work with this library. If you use another serde backend, you should test that it works first.

survived: > If you use another serde backend, you should test that it works first

If it works today, it doesn't mean it'll work tomorrow. Bincode worked with our lib before v0.2

survived: I expanded this section

maurges (approved):

PR #92, opened 2024-03-18
`key_share`: add `no_std` support

survived:
blocked by https://github.com/dfns/generic-ec/pull/25


drewstone (approved):

maurges (dismissed):

maurges:
Instead of making Display instances by hand, we could convert what we had with `thiserror` into https://docs.rs/displaydoc/

survived:
Nice! I'll try using it. Tho I'm not really a fan of deriving display impl from doc comments

survived:
I see that I can use `#[displaydoc("...")]` instead. I'll give it a shot

survived:
@maurges I changed the code to use `displaydoc` + `thiserror`. A bit prettier, but it did introduce a breaking change: the `HdError<E>` now only implements `Display` if `E: Display` (it used to implement display for regardless of `E`). However, I doubt it's an any significant breaking change...

key-share/src/lib.rs
```diff
@@ -468,41 +475,51 @@ impl<E: Curve> AsRef<CoreKeyShare<E>> for CoreKeyShare<E> {
 }
 
 /// Error indicating that key share is not valid
-#[derive(Debug, thiserror::Error)]
-#[error(transparent)]
-pub struct InvalidCoreShare(#[from] InvalidShareReason);
+#[derive(Debug, displaydoc::Display)]
+#[cfg_attr(feature = "std", derive(thiserror::Error))]
+#[displaydoc("invalid core key share")]
+pub struct InvalidCoreShare(#[cfg_attr(feature = "std", source)] InvalidShareReason);
```
maurges, line 481: Why change from transparent?

maurges: I guess because a transparent error implies an impossible Display implementation?

maurges:
> the `HdError<E>` now only implements `Display` if `E: Display`

It's silly that displaydoc crate doesn't allow us to request bounds by hand. But I agree that this is not really and important change

maurges (approved):

key-share/src/lib.rs
survived: Transparent error will generate Display implementation no matter what. We want `displaydoc` to do Display impl, not `thiserror`

PR #93, opened 2024-03-21
Fix serde issue

survived:
Turns out that `#[serde(flatten)]` is buggy, and, specifically, it does not preserve `is_human_readable` flag, which broke (de)serialization code and compatibility with old key shares.

See the issue for more details on the `serde` problem: https://github.com/serde-rs/serde/issues/2704

PR adds a workaround that manually reimplements `flatten` functionality.


key-share/src/serde_fix.rs
```diff
@@ -0,0 +1,42 @@
...
+    #[cfg(feature = "hd-wallets")]
+    #[cfg_attr(
+        feature = "serde",
+        serde(default),
+        serde(with = "As::<Option<crate::utils::HexOrBin>>")
+    )]
+    pub chain_code: Option<slip_10::ChainCode>,
+    #[serde(with = "As::<generic_ec::serde::Compact>")]
+    pub x: NonZero<SecretScalar<E>>,
+}
```
maurges: Can this struct instead have all its fields be references? This way there will be no allocations in serialization. I'm not sure if serde works correctly with references this way.

maurges: There would need to be a second struct with non-reference fields for deserialization though

survived: Yeah we can do this, but I didn't want to duplicate the struct and have serialization and deserialization in different places

maurges: On one hand, I don't see a problem with a little duplicating for a great cause. On the other, I think cloning here is cheap and acceptable. I leave the choice to you

survived: I just tried to duplicate the structs, and what I don't like in this approach is that it's difficult to track all the serde attributes we have. They need to be the same, so the serialization is compatible with deserialization, but when they're duplicated, it's difficult to check if there are any differences.

survived: I'll see if the macro helps

Cargo.toml
```diff
@@ -7,3 +7,6 @@ members = [
     "tests",
 ]
 
+[patch.crates-io.generic-ec]
+git = "https://github.com/dfns/generic-ec"
+branch = "better-compact"
```
survived: To be deleted

key-share/src/serde_fix.rs
survived: Yep it looks good (I had to do a small change in generic-ec tho)

survived:
Only in this PR I've noticed that we serialize scalar in `vss_setup` as not compact. It makes the key share bigger.

This is how compact things are serialized:

https://github.com/dfns/cggmp21/blob/f120531a851b00f1da6808f9961f00a32dc02a38/test-data/old-shares/v0.2/secp256k1-threshold-true-hd-false.json#L5-L11

this is how VSS setup is serialized:

https://github.com/dfns/cggmp21/blob/f120531a851b00f1da6808f9961f00a32dc02a38/test-data/old-shares/v0.2/secp256k1-threshold-true-hd-false.json#L12-L36

We can compress VSS setup by adding `generic_ec::serde::PreferCompact` that deserializes both compressed / not compressed, but serializes only the compact variant. Wdyt?

Cargo.toml
survived: https://github.com/dfns/generic-ec/pull/27

maurges:
> We can compress VSS setup by adding generic_ec::serde::PreferCompact that deserializes both compressed / not compressed, but serializes only the compact variant. Wdyt?

We should. I think it's what it should do by default, try both options on deserializing and always serialize as compact. You want to do it in this MR as well?

survived:
Better to do this in a new PR. We gotta release key-share crate, all versions are yanked

tests/tests/it/old_shares.rs
```diff
@@ -0,0 +1,45 @@
...
+
+fn deserialize_old_share_on_curve<E: generic_ec::Curve>(share: &include_dir::File) {
+    let ext = share
+        .path()
+        .extension()
+        .expect("share file name doesn't have extension")
+        .to_str()
+        .expect("key share ext is not valid utf8");
+
+    let _: cggmp21::IncompleteKeyShare<E> = match ext {
```
maurges, line 36: Might add checking if threshold and hd properties hold after deserialization. Though I think there's a low chance this goes wrong, so might be overkill. (I thought that was the reason you structured file names this way)

maurges:
Looks good, let's update generic-ec and go

tests/tests/it/old_shares.rs
survived: I just wanted to have every possible "kind" of key share. E.g. whether key share has or doesn't a chain code may result into a different kind of key share

survived: For validation - I assume that `Valid<_>` is good enough alone to validate a share

survived:
@maurges updated the generic-ec

tests/tests/it/old_shares.rs
maurges: I meant that, e.g., if the name has "hd-true", we verify that chaincode is present, and if "hd-false", verify that it's absent

tests/src/bin/precompute_shares.rs
```diff
@@ -11,13 +11,15 @@ use rand::{rngs::OsRng, CryptoRng, RngCore};
...
         Operation::GenPrimes => precompute_primes(),
     }
 }
 
 #[derive(Clone, Debug)]
 enum Operation {
     GenShares,
+    GenOldShares { out_dir: std::path::PathBuf },
     GenPrimes,
 }
```
maurges: Add clippy ignore? I see it fails

survived: Oh I ignored CI because of failing publish workflow... I thought only publish-* was failing :sweat_smile:

key-share/CHANGELOG.md
```diff
@@ -16,5 +27,7 @@
 [#82]: https://github.com/dfns/cggmp21/pull/82
 
 ## v0.1.0
+**YANKED**: this release is yanked because it had an issue with key share (de)serialization
+that was addressed in v0.2.2
```
maurges: The 1.0 was never released in the first place :D

survived: v0.1? it was https://crates.io/crates/key-share/0.1.0

survived: actually, v0.2.1 was never released...

tests/src/bin/precompute_shares.rs
survived: fixed

maurges (approved):

PR #101, opened 2024-05-24
Update deps

survived:
* Update deps (generic-ec, paillier-zk, slip-10) to latest
* Optimize key share verification and signing using new things from generic-ec


cggmp21-keygen/Cargo.toml
maurges, line 1: Shouldn't this version also be bumped? Or did we never release 0.2.0?

survived: We did not release it. I wanted to update generic-ec & round-based within one major bump

maurges (approved):

PR #104, opened 2024-07-22
`key_share`: add HD-related methods to `DirtyKeyInfo`

survived:


key-share/src/lib.rs
```diff
@@ -441,6 +441,33 @@ impl<E: Curve> DirtyCoreKeyShare<E> {
     }
 }
 
+#[cfg(feature = "hd-wallets")]
+impl<E: Curve> DirtyCoreKeyShare<E> {
+    /// Checks whether the key is HD-capable
+    pub fn is_hd_wallet(&self) -> bool {
+        (**self).is_hd_wallet()
+    }
```
maurges, line 449: Aren't these methods available automatically with DirtyCoreKeyShare implementing Deref into DirtyKeyInfo?

survived: They should be, but it’s still a breaking change to remove a method in public API (even if it’s still available via derefing)

survived: E.g. using DirtyCoreKeyShare::is_hd_wallet compiles before the removal, but doesn’t compile after the removal, derefing doesn’t work here

maurges: Since when do you care about breaking the API? Just bump the version :>

maurges (approved):

key-share/src/lib.rs
survived: Since always! I never was a fan of updating all the upstream deps, I only do that when there's no choice

PR #117, opened 2024-10-15
Add maintainers file

survived:


survived:
@maurges does it look good to populate in all repos?

maurges (approved):

maurges:
I think it's ok. I'm not sure I get the point though.

survived:
It's LFDT requirement that all repos have maintainers file which needs to list all the current maintainers, and document the process of becoming a maintainer (i.e. how someone from the outside can become a maintainer). The latter we have not formalized yet, so I'm only uploading a list.

PR #121, opened 2024-12-05
Update katex-header.html

survived:
PR updates katex-header.html file to be the same across all repos. Some of them were broken and caused double header menu displayed on docs.rs. I update all repos regardless of whether the katex header was broken, to sync all repos.


survived:
I'll fix CI issues in another PR. They are caused by CI using a new version of Rust compiler

maurges (approved):

PR #123, opened 2024-12-11
cggmp24 development

survived:
This is the branch in which we will work on updating implementation to the latest revision of CGGMP paper.

Note that `cggmp24/m` is protected branch, as well as `m` it requires a PR for any modification.

Upgrade plan:
1. Merge `paillier-zk` repo into the cggmp repo
1. Develop new ZK proofs needed for cggmp24
1. Update the cggmp implementation
1. Write migration guidelines
1. Rename repo into `cggmp24`
1. Release new library under `cggmp24` name
1. Legacy cggmp21 will get its own `cggmp21/m` protected branch reserved for security patches


survived:
@maurges does the plan looks okay?

maurges:
Sounds reasonable. We can even do the moving of zk-proofs here before starting to work on cggmp24.

I'm not sure we want to rename just the repo. I think the old repo name should stay in some form (didn't we have enough problems with branches disappearing, imagine a whole repo). Maybe we could rename the repo and make a fork under the old name, and immediately archive it

survived:
I don't anticipate any problems, as github will be redirecting old urls, so even if we have git dependencies (I don't think we do), they will get resolved with new url automatically

maurges:
> github will be redirecting old urls

It's not guaranteed it will do, it even says so when you rename a repo

survived:
> It's not guaranteed it will do, it even says so when you rename a repo

But it does it, e.g. https://github.com/LFDT-Lockness/slip-10 still redirects to the new repo.

The new release is going to be a breaking change, so it's okay even if git deps get broken

maurges:
It does redirect, but it's not guaranteed that it will do and do it forever

> The new release is going to be a breaking change, so it's okay even if git deps get broken

That's why I'm saying we should put an old version on the old URL

survived:
> > The new release is going to be a breaking change, so it's okay even if git deps get broken
> 
> That's why I'm saying we should put an old version on the old URL

I see. Yes it would be nice, but I want the repo name to reflect the crate name, and I don't want to keep the old one just to keep projects who use cggmp21 as git dependency intact.

survived:
@maurges just to add to this discussion: actually, we have already changed repo URL in the past, when we migrated from dfns org to LFDT, so why not doing it again? I don't see how anybody could reasonably expect stability from git dependencies...

survived:
`cggmp24/m` is now default branch in this repo

PR #127, opened 2024-12-20
Batch p2p messages in one round instead of flushing immediately

maurges:
I removed p2p message tracing from signing because I couldn't make sense of it. Ideas of how to track it are welcome

### DIFF
--- a/cggmp21-keygen/src/threshold.rs
+++ b/cggmp21-keygen/src/threshold.rs
@@ -277,15 +277,16 @@ where
         .await
         .map_err(IoError::send_message)?;
 
-    for j in utils::iter_peers(i, n) {
+    let messages = utils::iter_peers(i, n).map(|j| {
         let message = MsgRound2Uni {
             sigma: sigmas[usize::from(j)],
         };
-        outgoings
-            .send(Outgoing::p2p(j, Msg::Round2Uni(message)))
-            .await
-            .map_err(IoError::send_message)?;
-    }
+        Outgoing::p2p(j, Msg::Round2Uni(message))
+    });
+    outgoings
+        .send_all(&mut futures::stream::iter(messages.map(Ok)))
+        .await
+        .map_err(IoError::send_message)?;
     tracer.msg_sent();
 
     // Round 3
--- a/cggmp21-keygen/src/threshold.rs
+++ b/cggmp21-keygen/src/threshold.rs
@@ -284,7 +284,7 @@ where
         Outgoing::p2p(j, Msg::Round2Uni(message))
     });
     outgoings
-        .send_all(&mut futures::stream::iter(messages.map(Ok)))
+        .send_all(&mut futures_util::stream::iter(messages.map(Ok)))
         .await
         .map_err(IoError::send_message)?;
     tracer.msg_sent();
--- a/cggmp21-keygen/src/threshold.rs
+++ b/cggmp21-keygen/src/threshold.rs
@@ -271,7 +271,7 @@ where
 
     tracer.send_msg();
     outgoings
-        .send(Outgoing::broadcast(Msg::Round2Broad(
+        .feed(Outgoing::broadcast(Msg::Round2Broad(
             my_decommitment.clone(),
         )))
         .await
--- a/cggmp21/src/signing.rs
+++ b/cggmp21/src/signing.rs
@@ -715,13 +715,12 @@ where
         )
         .map_err(|e| Bug::PiEnc(BugSource::psi0, e))?;
 
-        tracer.send_msg();
         outgoings
-            .send(Outgoing::p2p(j, Msg::Round1b(MsgRound1b { psi0 })))
+            .feed(Outgoing::p2p(j, Msg::Round1b(MsgRound1b { psi0 })))
             .await
             .map_err(IoError::send_message)?;
-        tracer.msg_sent();
     }
+    outgoings.flush().await.map_err(IoError::send_message)?;
 
     // Round 2
     tracer.round_begins();
@@ -950,9 +949,8 @@ where
         .map_err(|e| Bug::PiLog(BugSource::psi_prime, e))?;
         runtime.yield_now().await;
 
-        tracer.send_msg();
         outgoings
-            .send(Outgoing::p2p(
+            .feed(Outgoing::p2p(
                 j,
                 Msg::Round2(MsgRound2 {
                     Gamma: Gamma_i,
@@ -967,8 +965,8 @@ where
             ))
             .await
             .map_err(IoError::send_message)?;
-        tracer.msg_sent();
     }
+    outgoings.flush().await.map_err(IoError::send_message)?;
 
     // Round 3
     tracer.round_begins();
@@ -1124,9 +1122,8 @@ where
         )
         .map_err(|e| Bug::PiLog(BugSource::psi_prime_prime, e))?;
 
-        tracer.send_msg();
         outgoings
-            .send(Outgoing::p2p(
+            .feed(Outgoing::p2p(
                 j,
                 Msg::Round3(MsgRound3 {
                     delta: delta_i,
@@ -1136,8 +1133,8 @@ where
             ))
             .await
             .map_err(IoError::send_message)?;
-        tracer.msg_sent();
     }
+    outgoings.flush().await.map_err(IoError::send_message)?;
 
     // Output
     tracer.named_round_begins("Presig output");
--- a/cggmp21/src/signing.rs
+++ b/cggmp21/src/signing.rs
@@ -687,7 +687,7 @@ where
 
     tracer.send_msg();
     outgoings
-        .send(Outgoing::broadcast(Msg::Round1a(MsgRound1a {
+        .feed(Outgoing::broadcast(Msg::Round1a(MsgRound1a {
             K: K_i.clone(),
             G: G_i.clone(),
         })))
--- a/cggmp21/src/signing.rs
+++ b/cggmp21/src/signing.rs
@@ -715,12 +715,16 @@ where
         )
         .map_err(|e| Bug::PiEnc(BugSource::psi0, e))?;
 
+        tracer.send_msg();
         outgoings
             .feed(Outgoing::p2p(j, Msg::Round1b(MsgRound1b { psi0 })))
             .await
             .map_err(IoError::send_message)?;
+        tracer.msg_sent();
     }
+    tracer.send_msg();
     outgoings.flush().await.map_err(IoError::send_message)?;
+    tracer.msg_sent();
 
     // Round 2
     tracer.round_begins();
@@ -949,6 +953,7 @@ where
         .map_err(|e| Bug::PiLog(BugSource::psi_prime, e))?;
         runtime.yield_now().await;
 
+        tracer.send_msg();
         outgoings
             .feed(Outgoing::p2p(
                 j,
@@ -965,8 +970,11 @@ where
             ))
             .await
             .map_err(IoError::send_message)?;
+        tracer.msg_sent();
     }
+    tracer.send_msg();
     outgoings.flush().await.map_err(IoError::send_message)?;
+    tracer.msg_sent();
 
     // Round 3
     tracer.round_begins();
@@ -1122,6 +1130,7 @@ where
         )
         .map_err(|e| Bug::PiLog(BugSource::psi_prime_prime, e))?;
 
+        tracer.send_msg();
         outgoings
             .feed(Outgoing::p2p(
                 j,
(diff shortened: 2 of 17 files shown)


maurges:
Welp, wasm build broke, and the error seems unrelated to my changes somehow

cggmp21-keygen/src/threshold.rs
```diff
@@ -277,15 +277,16 @@ where
...
-            .send(Outgoing::p2p(j, Msg::Round2Uni(message)))
-            .await
-            .map_err(IoError::send_message)?;
-    }
+        Outgoing::p2p(j, Msg::Round2Uni(message))
+    });
+    outgoings
+        .send_all(&mut futures::stream::iter(messages.map(Ok)))
+        .await
+        .map_err(IoError::send_message)?;
```
survived: Is it why you added `futures` dep, to do `futures::stream::iter`? You can use `futures-util` crate instead, i.e. [`futures_util::stream::iter`](https://docs.rs/futures-util/latest/futures_util/stream/fn.iter.html), `futures` crate just re-exports this function from `futures-util`

survived: Also, you could remove new (direct) dependency by using chain of [`outgoings.feed(msg).await`](https://docs.rs/futures-util/latest/futures_util/sink/trait.SinkExt.html#method.feed) inside of loop, followed by [`outgoings.flush().await`](https://docs.rs/futures-util/latest/futures_util/sink/trait.SinkExt.html#method.flush).

I would prefer this option, though using `futures_util::stream::iter` does have an advantage that we can't forget to call `.flush().await`.

survived: I see that in other places of the protocol, you are using `feed`/`flush` chain, so you can do that here as well and remove the dependency

cggmp21/src/signing.rs
```diff
@@ -715,13 +715,12 @@ where
...
 
-        tracer.send_msg();
         outgoings
-            .send(Outgoing::p2p(j, Msg::Round1b(MsgRound1b { psi0 })))
+            .feed(Outgoing::p2p(j, Msg::Round1b(MsgRound1b { psi0 })))
             .await
             .map_err(IoError::send_message)?;
-        tracer.msg_sent();
     }
+    outgoings.flush().await.map_err(IoError::send_message)?;
```
survived: You should keep `tracer.send_msg()`/`tracer.msg_sent()` so the time we spend on sending messages gets counted

survived: We can rename `send_msg`/`msg_sent` function names so the naming doesn't imply that a message has to be sent between these calls

survived: In fact, the only implementation of the trace trait we have does not assume that a single message actually gets sent between these calls. The time between these calls just get accounted as spent on sending messages, no matter how many were sent.

survived (commented):
You've only updated keygen/signing, are you going to update aux gen as well?

survived:
Also, would you be able to rebase PR to `cggmp24/m`? It's such a pain to sync changes from `m` to `cggmp24/m`...

cggmp21-keygen/src/threshold.rs
maurges: I use send_all where it doesn't change existing code too much, and feed+flush where send_all would. I think send_all is better, but I have a bias to not change existing lines too much

cggmp21/src/signing.rs
maurges: > In fact, the only implementation of the trace trait we have does not assume that a single message actually gets sent between these calls

Ugh, ok, that's a weird metric then. But I'll put the traces back in this case

survived: Measuring how much time we spend on sending messages in a weird metric to you?

maurges: That's a good metric. But we're not measuring it here, what we'll be measuring is time spent adding messages to a queue. I don't find it very useful, especially with a misleading name. Maybe it makes sense to add some trace to flushing though.

survived: We will be measuring time spent on adding to queue + time spent on flushing items from the queue. Two of them makes the time we spend on sending messages.

maurges: > Two of them makes

I expect that feeding messages makes them be processed in the background, io_uring/python asyncio style. So the two of the measurements is not greater than the time spent on sending messages

We can talk about "effective time spent" or something, which might be what we really care about, and then it will be useful again. I'm adding the traces anyway

survived: > I expect that feeding messages makes them be processed in the background

`feed` does the same thing as `send`, it just doesn't do `flush` in the end

survived: There's no magical background processors for sinks. Either sink itself supports that or not. If it doesn't, then messages are sent within protocol.

Sink can indeed send messages to other task to do sending, but then caller should not expect `Tracer` to be able to measure time we spend on sending messages, as protocol does not get to observe this.

survived: > I expect that feeding messages makes them be processed in the background, io_uring/python asyncio style

You shouldn't expect that, it's not python, nor was it inspired by io_uring :smiley:

cggmp21-keygen/src/threshold.rs
```diff
@@ -277,15 +277,16 @@ where
         .await
         .map_err(IoError::send_message)?;
```
survived, line 278: Can't leave a comment above, but the send there should be `feed` so it doesn't invoke `flush` until all messages to be sent are in the sink

survived: https://github.com/LFDT-Lockness/cggmp21/blob/44695389cf1333672c6907f4f298bcd22b0adba3/cggmp21-keygen/src/threshold.rs#L273-L278

cggmp21/src/signing.rs
```diff
@@ -717,11 +717,14 @@ where
 
         tracer.send_msg();
         outgoings
-            .send(Outgoing::p2p(j, Msg::Round1b(MsgRound1b { psi0 })))
+            .feed(Outgoing::p2p(j, Msg::Round1b(MsgRound1b { psi0 })))
             .await
             .map_err(IoError::send_message)?;
         tracer.msg_sent();
```
survived, line 723: In this round, we send a broadcast message, it should use `feed` as well so we don't flush it before all messages are in the sink

survived: https://github.com/LFDT-Lockness/cggmp21/blob/44695389cf1333672c6907f4f298bcd22b0adba3/cggmp21/src/signing.rs#L688-L696

survived (commented):
It might be a good idea to modify sink in the tests so it actually waits until `flush` is called to send all messages. In this way, if we forget to call `flush` in the protocol, the tests get deadlocked and we notice that.

I think current behavior of sink in the tests is sending messages to other parties as soon as it's given to the sink

maurges:
> It might be a good idea to modify sink in the tests so it actually waits until flush is called

Good idea. As I see the sink is defined in round-based, right?

cggmp21/src/signing.rs
maurges: > There's no magical background processors for sinks. Either sink itself supports that or not. If it doesn't, then messages are sent within protocol

Exactly. And here we're writing code for any sink, so a sink that does background processing will show up weird in the tracer. 

> caller should not expect Tracer to be able to measure time we spend on sending messages

That's exactly what I'm telling you: depending on the sink implementation, these sending timings are completely wrong

survived:
> Good idea. As I see the sink is defined in round-based, right?

Yes. I was thinking that we could do a wrapper sink that would do what we want. 

maybe it's sufficient to have a one separate test that modifies a default sink provided by round-based and runs all the protocols to make sure `flush` is always called.

cggmp21/src/signing.rs
survived: Depending on sink implementation, caller should choose appropriate tracer implementation. If sink doesn't actually send messages but rather sends them to the other task, then tracer should not measure what time it takes for `send` calls

tests/src/lib.rs
```diff
@@ -4,6 +4,81 @@ use generic_ec::Curve;
...
+
+    fn poll_flush(
+        mut self: std::pin::Pin<&mut Self>,
+        cx: &mut std::task::Context<'_>,
+    ) -> std::task::Poll<Result<(), Self::Error>> {
+        while !self.messages.is_empty() {
+            let mut projection = self.as_mut().project();
+            let mut inner = projection.inner;
+            match inner.as_mut().poll_ready(cx) {
+                std::task::Poll::Pending => return std::task::Poll::Pending,
```
survived: nitpick: you can use [`std::task::ready!` macro](https://doc.rust-lang.org/stable/std/task/macro.ready.html) instead

tests/src/lib.rs
```diff
@@ -4,6 +4,81 @@ use generic_ec::Curve;
...
+                std::task::Poll::Ready(Ok(())) => {
+                    if let Some(item) = projection.messages.pop() {
+                        inner.as_mut().start_send(item)?;
+                    }
```
survived: I guess it doesn't really matter in our usage, but it's also a bit unfortunate that buffered sink reverses the order of messages

survived: Were you considering using [`VecDeque`](https://doc.rust-lang.org/stable/std/collections/struct.VecDeque.html) instead?

tests/src/lib.rs
```diff
@@ -4,6 +4,81 @@ use generic_ec::Curve;
...
+    party: round_based::MpcParty<M, D, R>,
+) -> round_based::MpcParty<M, BufferedDelivery<M, D>, R>
+where
+    M: Unpin,
+    D: round_based::Delivery<M>,
+    R: round_based::runtime::AsyncRuntime,
+{
+    let round_based::MpcParty {
+        delivery, runtime, ..
+    } = party;
```
survived: Hmm, maybe it's better if we add `MpcParty::map_delivery` method?

survived: This code might shoot us in the foot if we add some important fields to `MpcParty` in the future

tests/src/lib.rs
```diff
@@ -4,6 +4,81 @@ use generic_ec::Curve;
...
+    }
+}
+
+pub fn buffer_outgoing<M, D, R>(
+    party: round_based::MpcParty<M, D, R>,
+) -> round_based::MpcParty<M, BufferedDelivery<M, D>, R>
+where
+    M: Unpin,
+    D: round_based::Delivery<M>,
+    R: round_based::runtime::AsyncRuntime,
```
survived, line 77: I think it's not obvious from the code, why we do bufferization in the tests at all. Worth adding a comment?

tests/src/lib.rs
maurges: Yes, I think we should

tests/src/lib.rs
maurges: Makes sense to use, I got to the current implementation through a wrong idea of draining the whole vector, so there was no need to deque

tests/src/lib.rs
survived: I think it's better to do within this PR? Otherwise it's easy to forget

tests/src/lib.rs
```diff
@@ -4,6 +4,85 @@ use generic_ec::Curve;
...
+        self.project().inner.poll_close(cx)
+    }
+}
+
+/// Map a party to add explicit buffering to outgoing messages in case the
+/// underlying sink doesn't do buffering. This is useful to test that we don't
+/// forget to flush the sinks in our protocol implementations.
+pub fn buffer_outgoing<M, D, R>(
+    party: round_based::MpcParty<M, D, R>,
+) -> round_based::MpcParty<M, BufferedDelivery<M, D>, R>
```
survived: I dunno, sounds a bit obscure. Maybe this?

 ```suggestion
/// Modifies `Delivery` implementation of the `MpcParty` so sink of outgoing messages
/// buffers messages to be sent until `flush` is called.
///
/// cggmp21 protocol implementation often intentionally delay the moment the `flush` is called
/// to optimize network delays and let delivery implementation parallelize message sending.
/// However, we might accidentally forget to call `flush()` in the end of the round. By calling this
/// method in the tests and buffering all outgoing messages until `flush` is called we make sure
/// that we did not forget to call `flush`: otherwise tests will deadlock. 
pub fn buffer_outgoing<M, D, R>(
    party: round_based::MpcParty<M, D, R>,
) -> round_based::MpcParty<M, BufferedDelivery<M, D>, R>
```

tests/src/lib.rs
maurges: You mean make a release of round-based before merging this one? Ok, I can do that

survived: Yeah a minor release

survived (approved):

PR #128, opened 2025-01-08
Update the spec

survived:
The spec is currently in state of chaos:

* `main.tex` is poorly formatted as several people were writing and rewriting it with different style preferences. E.g. different part of the spec use 2 space indent, while others use 4 space
* Large parts of the spec are commented out, which should probably be deleted or addressed (if it's kind of todo remark)
* Sometimes different notations are used
* Sometimes spec doesn't reflect what happens in the code

The idea of this PR is to update the spec sources to have uniform formatting, make sources easier to read, preferably use a formatting tool and integrate it into CI. Also, unify notations and sync specs and the code where they are not synced.

In addition to that, I also want to make types of the messages received during the protocols explicit, e.g. 
* instead of "Upon receiving $\psi_j$ from all parties, ..."
* say this: "Upon receiving $\psi_j \in \mathbb F_q$ from all parties, ..."

which explicitly says that the implementation does need to check that we received integer in $\mathbb F_q$.

Progress:
- [x] Aux gen
- [x] DKG (threshold & non-threshold)
- [x] Signing
- [ ] Unify formatting & find fmt tool

### DIFF
--- a/paillier-zk/src/no_small_factor.rs
+++ b/paillier-zk/src/no_small_factor.rs
@@ -30,8 +30,8 @@
 //!
 //! let aux: p::Aux = pregenerated::verifier_aux();
 //! let security = p::SecurityParams {
-//!     l: 4,
-//!     epsilon: 128,
+//!     l: 256,
+//!     epsilon: 230,
 //!     q: (Integer::ONE << 128_u32).complete(),
 //! };
 //!
--- a/shell.nix
+++ b/shell.nix
@@ -7,7 +7,7 @@ let
   # Latex
   tex = (pkgs.texlive.combine {
     inherit (pkgs.texlive) scheme-small
-      collection-mathscience preprint amsmath;
+      collection-mathscience preprint amsmath enumitem placeins;
   });
 
 in pkgs.stdenv.mkDerivation {
--- a/shell.nix
+++ b/shell.nix
@@ -1,16 +1,9 @@
-let # Rust
-  pkgs = import <nixpkgs> { overlays = [ rustOverlay ]; };
-  lib = pkgs.lib;
+# shell for compiling latex spec
+
+let
+  pkgs = import <nixpkgs> {};
   isDarwin = pkgs.hostPlatform.isDarwin;
 
-  rustVersion = "1.75.0";
-  rustOverlay = import (builtins.fetchTarball "https://github.com/oxalica/rust-overlay/archive/master.tar.gz");
-
-  rust = pkgs.rust-bin.stable.${rustVersion}.default.override {
-    extensions = [
-      "rust-src" # for rust-analyzer
-    ];
-  };
   # Latex
   tex = (pkgs.texlive.combine {
     inherit (pkgs.texlive) scheme-small
@@ -20,7 +13,7 @@ let # Rust
 in pkgs.stdenv.mkDerivation {
   name = "signers-env";
   nativeBuildInputs = [
-    rust pkgs.rust-analyzer tex pkgs.gnum4
+    tex
   ];
-  buildInputs = lib.optionals isDarwin [pkgs.darwin.apple_sdk.frameworks.Security];
+  buildInputs = pkgs.lib.optionals isDarwin [pkgs.darwin.apple_sdk.frameworks.Security];
 }
--- a/shell.nix
+++ b/shell.nix
@@ -2,7 +2,6 @@
 
 let
   pkgs = import <nixpkgs> {};
-  isDarwin = pkgs.hostPlatform.isDarwin;
 
   # Latex
   tex = (pkgs.texlive.combine {
@@ -15,5 +14,4 @@ in pkgs.stdenv.mkDerivation {
   nativeBuildInputs = [
     tex
   ];
-  buildInputs = pkgs.lib.optionals isDarwin [pkgs.darwin.apple_sdk.frameworks.Security];
 }
--- a/spec/main.tex
+++ b/spec/main.tex
@@ -1988,7 +1988,7 @@ The signing protocol has two parts: one that takes the output from the presignat
 
 \item
 \begin{inlineAlgorithm}
-\algoName{$\fn{combine\_partial\_signatures(Y, \{(r_i, \sigma_i)\}_{i=0}^{n-1}, m)} \to (r, \sigma)$}
+\algoName{$\fn{combine\_partial\_signatures}(Y, \{(r_i, \sigma_i)\}_{i=0}^{n-1}, m) \to (r, \sigma)$}
 \algoInputsList{
     \item public key $Y \in \E$,
     \item partial signatures $\{(r_i, \sigma_i)\}_{i=0}^{n-1} \in (\Z_q, \Z_q)^n$,
--- a/spec/main.tex
+++ b/spec/main.tex
@@ -1265,7 +1265,7 @@ In all the cases where this proof is used in the protocol, the prover knows the 
 
 \subsection{$\proof{fac}$: No Small Factor Proof}
 The prover and verifier agree on shared state $\state$, auxiliary data $R_j=(N_j, s_j, t_j)$ with $s_j, t_j \in {\mathbb Z}_{N_j}^*$, and a security level~$L$. 
-For this proof, the prover and verifier have common input~$N_i$, and the prover additionally has primes $2^\ell < p, q < \pm 2^\ell \cdot \sqrt{N_i}$ with $N_i=pq$. 
+For this proof, the prover and verifier have common input~$N_i$, and the prover additionally has primes $2^\ell < p, q < 2^\ell \cdot \sqrt{N_i}$ with $N_i=pq$. 
 In all the cases where this proof is used in the protocol, the verifier knows the factorization of $N_j$ (and hence knows~$\sk_j$).
 
 %  \item Proof guarantees: $p_0, q_0 > 2^\ell$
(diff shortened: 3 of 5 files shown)


survived:
One thing I didn't figure out how to do nicely, is how to specify domain of the received proof. E.g. in aux gen protocol round 3:

* We receive $(N_j, s_j, t_j, \hat{\psi}_j, \rho_j, u_j)$
* where $\hat{\psi}\_j$ is prm proof, that consists of two lists of integers $(\\{A\_i\\}^m\_{i=1}, \\{z\_i\\}^m\_{i=1}) \in (\mathbb{Z}\_N^m, \mathbb{Z}\_{N}^m)$ 
* For rest of args it's easy to specify their domain: $(N_j, s_j, t_j, \rho_j, u_j) \in (\mathbb{Z}, \mathbb{Z}_N, \mathbb{Z}_N, \\{0,1\\}^\kappa, \\{0,1\\}^\kappa)$

For now I just assume the proof validates domain of the input arguments, but in fact, in the spec they do not do that.

spec/main.tex
```diff
@@ -886,13 +887,15 @@ \subsection{Provisioning Protocol}
 
       \item For $j \neq i$:
         \begin{itemize}
+          \item Assert $(N_j, s_j, t_j, \rho_j, u_j) \? \in
+            (\Z, \Z, \Z, \Bit^\kappa, \Bit^\kappa)$
```
survived: Surprisingly, we never require anywhere that $s_j < N_j$ and $t_j < N_j$. Probably it's redundant?

spec/main.tex
```diff
@@ -886,13 +887,15 @@ \subsection{Provisioning Protocol}
...
           \item Assert $V_j =
-            H(\Encode{hash\_com}(\sid, n, j, N_j, s_j, t_j, \hat{\psi}_j, \rho_j, u_j))$.
+            H(\Encode{hash\_com}(\sid, j, N_j, s_j, t_j, \hat{\psi}_j, \rho_j, u_j))$.
           \item Assert $N_j$ is at least $8 \cdot \kappa - 1$ bits in length
```
survived: We only set a lower limit at the $N_j$, but shouldn't we set upper limit as well? I'm thinking that if adversary can set a large $N$, then protocol might significantly slow down.

survived (commented):
A few thoughts after having gone through aux gen protocol

paillier-zk/src/no_small_factor.rs
```diff
@@ -30,8 +30,8 @@
 //!
 //! let aux: p::Aux = pregenerated::verifier_aux();
 //! let security = p::SecurityParams {
-//!     l: 4,
-//!     epsilon: 128,
+//!     l: 256,
+//!     epsilon: 230,
```
survived, line 34: Note: previous params set has failed on CI: https://github.com/LFDT-Lockness/cggmp21/actions/runs/12690901369/job/35372765003

survived:
@maurges can you take a look before I start rewriting all proofs?

shell.nix
```diff
@@ -1,26 +1,19 @@
...
   });
 
 in pkgs.stdenv.mkDerivation {
   name = "signers-env";
   nativeBuildInputs = [
-    rust pkgs.rust-analyzer tex pkgs.gnum4
+    tex
   ];
-  buildInputs = lib.optionals isDarwin [pkgs.darwin.apple_sdk.frameworks.Security];
+  buildInputs = pkgs.lib.optionals isDarwin [pkgs.darwin.apple_sdk.frameworks.Security];
```
maurges: We don't build the rust code in nix-shell anymore, right? This line was used for rust builds, but it seems it's no longer necessary

survived: I had no idea what it does, so I left it. I will remove it then

maurges:
> @maurges can you take a look before I start rewriting all proofs?

Looks good, go ahead

survived:
Can you check this out? @maurges I've updated all proofs, next I want to update the protocol specs, so they also use `algorithmic`, but I want to let you review the proofs first, so you could see changes in the protocol that I had to make due to changes in the proofs

Overall, there's no significant changes, I just unified notation, and sometimes rearranged order of arguments. I also reviewed again all ZK proofs implementations to be aligned with the spec, all looks good.

spec/main.tex
maurges: Maybe because all operation are already modulo reduced? (I didn't check the fact I just stated)

spec/main.tex
maurges: Well, in this case we literally follow the check mandated by the original paper. I agree with you, adding such a check makes sense

survived: Yes, paper doesn't care if someone tries to DoS us with gigabyte-ish integers...

survived:
@maurges the final touch: I updated all protocols to use `algorithmic` for consistency. That's all I wanted to do within this PR

survived:
well, not counting automatic formatting, but turns out that easier to migrate to typst...

survived:
(@maurges it's ready for the final review)

maurges:
Screenshot 2025-02-21 at 14 10 18

In provisioning protocol: don't see this in the code, and don't see it being used in the description neither

maurges:
Screenshot 2025-02-21 at 14 13 40

A line in n-of-n keygen got fucked up a bit. May be my pdflatex bug

maurges:
Screenshot 2025-02-21 at 15 26 06

n out of n presigning round 1: in the other places you write `j \in [n] \minus {i}`. You use this old notation in a couple of other places in this protocol

spec/main.tex
```diff
@@ -1100,158 +1809,203 @@ \subsubsection{Non-Threshold ($n$-out-of-$n$) Presigning} \label{section:orig-pr
...
-                and $F_{j,i} = \enc^\crt_{\sk_i}(-\beta_{i,j}; r_{i,j})$.
-            \item Compute $\hat{D}_{j,i} = (x_i \odot K_j) \oplus \enci{j}(-\hat{\beta}_{i,j}; \hat{s}_{i,j})$
-                and $\hat{F}_{j,i} = \enc^\crt_{\sk_i}(-\hat{\beta}_{i,j}; \hat{r}_{i,j})$.
-
-            \item Compute $\psi_{j,i} = \proveni{aff\mbox{-}g}^{\E,L}((\sid, i), R_j, (N_j, N_i, K_j, D_{j,i}, F_{j,i}, \Gamma_i); (\gamma_i, -\beta_{i,j}, s_{i,j}, r_{i,j}))$ and
-            $\hat{\psi}_{j,i} = \proveni{aff\mbox{-}g}^{\E,L}((\sid, i), R_j, (N_j, N_i, K_j, \hat{D}_{j,i}, \hat{F}_{j,i}, X_i); (x_i, -\hat{\beta}_{i,j}, \hat{s}_{i,j}, \hat{r}_{i,j}))$. 
-            
+    \item[\textbf{Round 2.}] \
+    \begin{algorithmic}[1]
+        \State Receive $(K_j, G_j, \psi^0_{i,j})$ from all parties
```
maurges: There are no domains in received data anywhere in this algorithm

spec/main.tex
```diff
@@ -1100,158 +1809,203 @@ \subsubsection{Non-Threshold ($n$-out-of-$n$) Presigning} \label{section:orig-pr
...
-        
-        \item If shares are Shamir secret shares of the private key: 
+    \item[\bf Input.] \
+    \begin{itemize}[noitemsep,topsep=0pt]
+        \item size of signing set $t \ge 2$,
+        \item party index $i \in [t]$,
+        \item injective index map $S : [t] \to [n]$ that maps signers indicies at signing to their indicies at key generation, i.e. $S(i)$ is the index that $P_i$ had at key-generation time,
+        \item key share $K_{S(i)}$,
+        \item context separation string $\sid \in \Byte^*$,
+        \item additive shift $\var{shift} \in \Z_q$ for HD child derivation. $\var{shift} = 0$ disables HD derivation. Deriving additive shift is up to specific standard of HD-wallets, e.g. see \cite{bip32} or \cite{slip10} \label{deriving-shift-note}
```
maurges: The phrasing about "is up to a specific standard..." is a bit confusing. I would expand to something like "The procedure for deriving this shift from a human representation of an HD-wallet is out of scope for this algorithm; for specific standards additive shift derivation see ...". And probably move it to a footnote.

Also, I don't see the `\label` having any effect in the pdf

spec/main.tex
```diff
@@ -1100,158 +1809,203 @@ \subsubsection{Non-Threshold ($n$-out-of-$n$) Presigning} \label{section:orig-pr
...
         \end{itemize}
-    \end{itemize}
 
-    \item[\bf Step 2.] {\bf (HD-wallets)} If HD-wallets support enabled:
-    \begin{itemize}
-      \item Set $X_0 := X_0 + \text{shift} \cdot G$
-      \item If $i = 0$, set $x_0 := x_0 + \text{shift}$
-      \item Note: output signature will be valid for public key $Y + \text{shift} \cdot G$
-    \end{itemize}
+        \State Set $\sk_i = \sk'_{S(i)}$ and $\vec R = (R'_{S(j)})_{j \in [n]}$
```
maurges: Pretty sure it should be `j \in [t]` (in the next step as well)

spec/main.tex
```diff
@@ -1100,158 +1809,203 @@ \subsubsection{Non-Threshold ($n$-out-of-$n$) Presigning} \label{section:orig-pr
...
+        \If{shares are additive\footnote{In this case we have $t=n$.} shares of the private key}
+            \State Set $x_{i} = x'_{S(i)}$, $\vec X = (X'_{S(j)})_{j \in [n]}$
+        \EndIf
+        \If{shares are Shamir secret shares of the private key}
```
maurges: I think this should rather be `\elseif`, does algorithmic support it?

spec/main.tex
```diff
@@ -1100,158 +1809,203 @@ \subsubsection{Non-Threshold ($n$-out-of-$n$) Presigning} \label{section:orig-pr
...
+    \item public key $Y \in \E$,
+    \item partial signatures $\{(r_i, \sigma_i)\}_{i=0}^{n-1} \in (\Z_q, \Z_q)^n$,
+    \item a hashed message $m \in \Z_q$
+}
+\algoOutputs{signature $(r, \sigma) \in (\Z_q, \Z_q)$}
+\begin{algorithmic}[1]
+\State Assert $r_0 \? = r_0 \? = \dots \? = r_{n-1}$
+\State Set $r := r_0$
+\State Let $\sigma = \sum_j \sigma_j \bmod q$
+\State Verify that $(r, \sigma)$ is a valid ECDSA signature on hashed message $m$ with respect to public key $Y$, abort if it's not
```
maurges: `s/Verify that (.*), abort if it's not/Assert that \1`?

spec/main.tex
survived: That's from the old spec, I didn't change the phrasing or `\label`, happy to change this

survived: > a human representation of an HD-wallet

Not sure this phrase makes sense though, I imagine many humans would argue that a list of indexes is not human-representation at all

survived: I'd keep this:

> The procedure for deriving the shift is out of scope for this algorithm; for specific standards for shift derivation see ...

But it says basically the same thing to me as what we have now

spec/main.tex
survived: I tried exactly this, `\Elseif`, it gave me an error, I didn't want to mess with it, so I made a bunch of `\If`

survived:
> Screenshot 2025-02-21 at 14 13 40
> 
> A line in n-of-n keygen got fucked up a bit. May be my pdflatex bug

Looks fine on my end:

image

spec/main.tex
maurges: I like your option. I don't think it's the same thing, this one explicitly says "out of scope"

maurges:
Looks good. I'd like to have one rephrasing in discussion here https://github.com/LFDT-Lockness/cggmp21/pull/128#pullrequestreview-2637278240 and that's it

spec/main.tex
survived: Moved to #136

maurges (approved):

PR #129, opened 2025-01-08
chore: update actions

ryjones:
actions/upload-artifact and actions/download-artifact have been deprecated and need updated.

https://github.blog/changelog/2024-04-16-deprecation-notice-v3-of-the-artifact-actions/

### DIFF
--- a/.github/workflows/deps.yml
+++ b/.github/workflows/deps.yml
@@ -19,11 +19,11 @@ jobs:
       contents: read
       pull-requests: write
     steps:
-    - uses: actions/checkout@v3
+    - uses: actions/checkout@v4
       with:
         path: base_branch
         ref: ${{ github.base_ref }}
-    - uses: actions/checkout@v3
+    - uses: actions/checkout@v4
       with:
         path: pr_branch
     - name: Retrieve base branch deps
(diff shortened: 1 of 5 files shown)


survived (approved):

survived:
Thanks for the PR! We have a policy that prevents us from merging unsigned commits @ryjones. Can you sign them? Or alternatively I can re-do PR with my commits

ryjones:
@survived I can but it would break your commit signature, so please go ahead

ryjones:
@survived I added your changes to my commit

survived:
It's strange that my review was not dismissed after you pushed a new commit despite the protection settings of the branch. Probably a github bug?

survived:
The diff is the same as approved one, so maybe github didn't treat the new commit as something that should dismiss the review approval

ryjones:
@survived I updated [the ruleset](https://github.com/LFDT
Screenshot 2025-01-10 at 08 35 21
-Lockness/cggmp21/settings/rules/2940679)

survived:
I don't think it will work actually, all it do (from my experience) is that if I do the most recent push, I can't approve PR

survived:
in this case, you did the most recent push, I did the review, this policy does not apply (I also think it has confusing name)

PR #137, opened 2025-02-27
Trusted-dealer can generate random share preimages

maurges:
Add new functions to `key_share::trusted_dealer::TrustedDealerBuilder`: `generate_shares_at` and `generate_shares_at_random`

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -35,7 +35,7 @@ futures-util = { version = "0.3", default-features = false }
 
 anyhow = "1"
 thiserror = "1"
-displaydoc = { version = "0.2", default-features = false }
+displaydoc = { version = "0.2.5", default-features = false }
 
 serde = { version = "1", default-features = false }
 serde_with = { version = "2", default-features = false }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,7 +25,7 @@ udigest = { version = "0.2.1", default-features = false }
 digest = { version = "0.10", default-features = false }
 sha2 = { version = "0.10", default-features = false }
 
-rand = "0.8"
+rand = { version = "0.8", default-features = false }
 rand_core = { version = "0.6", default-features = false }
 rand_hash = { version = "0.1" }
 rand_dev = "0.1"
--- a/cggmp21/src/trusted_dealer.rs
+++ b/cggmp21/src/trusted_dealer.rs
@@ -147,6 +147,8 @@ impl<E: Curve, L: SecurityLevel> TrustedDealerBuilder<E, L> {
 
     /// Generates [`IncompleteKeyShare`]s
     ///
+    /// For Shamir secret sharing, it's shared at points `1` to `n`
+    ///
     /// Returns error if provided inputs are invalid, or if internal
     /// error has occurred.
     pub fn generate_core_shares(
@@ -169,11 +171,10 @@ impl<E: Curve, L: SecurityLevel> TrustedDealerBuilder<E, L> {
         self,
         rng: &mut (impl RngCore + CryptoRng),
     ) -> Result<Vec<KeyShare<E, L>>, TrustedDealerError> {
-        let preimages = (1..=self.n)
-            .map(|i| generic_ec::NonZero::from_scalar(generic_ec::Scalar::from(i)))
-            .collect::<Option<Vec<_>>>()
-            .ok_or(Reason::DeriveKeyShareIndex)?;
-        self.generate_shares_at(preimages, rng)
+        self.generate_shares_at_internal(
+            key_share::trusted_dealer::TrustedDealerBuilder::generate_shares,
+            rng,
+        )
     }
 
     /// Generates [`KeyShare`]s shared at preimages provided. Each share is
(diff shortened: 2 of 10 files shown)


Cargo.toml
```diff
@@ -25,7 +25,7 @@ udigest = { version = "0.2.1", default-features = false }
 digest = { version = "0.10", default-features = false }
 sha2 = { version = "0.10", default-features = false }
 
-rand = "0.8"
+rand = { version = "0.8", default-features = false }
```
survived, line 28: Why? This enables default features in all subcrates that depend on `rand`. We need default features in all of them?

survived: Maybe better specify the exact features you need?

key-share/src/trusted_dealer.rs
```diff
@@ -97,18 +97,62 @@ impl<E: Curve> TrustedDealerBuilder<E> {
...
+    pub fn generate_shares_at_random(
+        self,
+        rng: &mut (impl rand_core::RngCore + rand_core::CryptoRng),
+    ) -> Result<Vec<CoreKeyShare<E>>, TrustedDealerError> {
+        let key_shares_indexes =
+            rand::seq::index::sample(rng, usize::from(u16::MAX - 1), usize::from(self.n))
+                .iter()
+                .map(|i| generic_ec::NonZero::from_scalar(Scalar::from(i + 1)))
+                .collect::<Option<Vec<_>>>()
+                .ok_or(Reason::DeriveKeyShareIndex)?;
```
survived: I'm super confused. Why not just generate `n` random scalars? Why do you want that all scalars are < u16::MAX ?

key-share/src/trusted_dealer.rs
```diff
@@ -97,18 +97,62 @@ impl<E: Curve> TrustedDealerBuilder<E> {
...
+    ///
+    /// Preimages are ignored for additive key shares.
+    ///
+    /// Returns error if provided inputs are invalid, or if internal
+    /// error has occurred.
+    pub fn generate_shares_at(
+        self,
+        preimages: Vec<NonZero<Scalar<E>>>,
+        rng: &mut (impl rand_core::RngCore + rand_core::CryptoRng),
+    ) -> Result<Vec<CoreKeyShare<E>>, TrustedDealerError> {
```
survived, line 151: This function should return an error if `preimages.len() != n`. Currently it doesn't check length of `preimages`

Cargo.toml
maurges: Are you reading this backwards? I disabled default features in rand; because now rand is used in key-share, but by default rand requires `std`. So I disable default features to disable it

key-share/src/trusted_dealer.rs
maurges: Hmm, it seems you're right, we don't need them to be related to u16 in any way. A brain fart on my part

Cargo.toml
survived: Yes I was reading it backwards 🤦‍♂️

key-share/src/trusted_dealer.rs
```diff
@@ -97,18 +97,73 @@ impl<E: Curve> TrustedDealerBuilder<E> {
...
+                let point = generic_ec::NonZero::<Scalar<E>>::random(rng);
+                if !points.contains(&point) {
+                    points.push(point);
+                    continue 'each_point;
+                }
+            }
+            // if we did not continue in inner loop, it means we couldn't
+            // generate a distinct scalar
+            return Err(Reason::BadRandom.into());
+        }
```
survived: Isn't it an overkill? If you just generate `n` random scalars, they will be all distinct with overwhelming probability. 

Probability that there will be at least two equal scalars among `n` random scalars is:

$$P(n) \approx \frac{n^2}{2q}$$

Given that `n` is u16 and assuming $q \approx 2^{256}$, the max probability is:

$$P(2^{16}) \approx \frac{2^{32}}{2^{257}} = 2^{-225}$$

which is not something we should worry about

key-share/src/trusted_dealer.rs
```diff
@@ -97,18 +97,73 @@ impl<E: Curve> TrustedDealerBuilder<E> {
...
+    ///
+    /// Preimages are ignored for additive key shares.
+    ///
+    /// Returns error if provided inputs are invalid, or if internal
+    /// error has occurred.
+    pub fn generate_shares_at(
+        self,
+        preimages: Vec<NonZero<Scalar<E>>>,
+        rng: &mut (impl rand_core::RngCore + rand_core::CryptoRng),
+    ) -> Result<Vec<CoreKeyShare<E>>, TrustedDealerError> {
```
survived, line 151: In this function it might make sense to add a check that all preimages are distinct to protect from invalid input, but I think the key share validation logic already does check that, so no need to do that here

maurges: It doesn't check it explicitly, but if we have duplicate preimages, we will divide by zero in one of lagrange interpolations, which will be caught

cggmp21/src/trusted_dealer.rs
survived, line 1: cggmp21 crate has changes, do you want to bump its version as well?

survived:
@maurges looks like in this PR, we only need to bump a lib version and it's good to go

survived (approved):

PR #145, opened 2025-06-04
Fix sampling from wrong range 

### LINKED ISSUE #144: SigningError(Bug(PaillierEnc(neg_beta_ij_enc)))
Hi,

We encountered the following error, when attempting to sign a message `hello` with a 2-of-3 setup.

```
SigningError(Bug(PaillierEnc(neg_beta_ij_enc)))
```

Could it be a bug somewhere? 

I've also attached the key shares [here](https://github.com/user-attachments/files/20519974/keys.txt).

Thanks,
Yulong

survived:
There was sampling from $N_i$ whereas spec says to sample from $N_j$. That was passed undetected until we updated `fast-paillier` to reject $x \ge N$ when checking $x \in \mathbb{Z}^*_N$, which now makes the protocol randomly abort when $N_j < N_i$ and sampled $x  > N_j$.

Fixes #144

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -298,7 +298,7 @@ checksum = "baf1de4339761588bc0619e3cbc0120ee582ebb74b53b4efbf79117bd2da40fd"
 
 [[package]]
 name = "cggmp21"
-version = "0.6.0"
+version = "0.6.1"
 dependencies = [
  "cggmp21-keygen",
  "digest",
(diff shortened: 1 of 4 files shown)


survived:
@maurges can you quickly take a look please? Currently, cggmp21 is broken when used with latest `fast-paillier`

maurges (approved):

PR #147, opened 2025-06-05
Update aux generation & signing protocols to latest revision

survived:
Originally, I wanted to update the aux protocol only, but the signing just doesn't work, and because of that so many tests and the whole CI were broken, thus in this PR I'm updating both aux gen & signing protocols.

Changelog:
* Update aux gen & signing protocols to match cggmp24 latest revision
* Remove non-threshold key refresh protocol (such a burden)
* Update key share structure
  * Core key share (a.k.a. `IncompleteKeyShare`) in not changed
  * Aux data is changed: we now have two different keys, Paillier key and Pedersen key per each signer.
  * Provide migration tool and instructions in `cggmp24::key_share::compat`
* Update SecurityLevel macro to match the latest revision and refactor it (something I wanted to do so long):
  * Add `RSA_PRIME_BITLEN` and `RSA_PUBKEY_BITLEN` params
  * Remove parameter `Q` (it's gone from the paper)
  * Remove parameter `M` (that's actually hardcoded into `cggmp24::security_level::M`)
  * Rename:
    * `SecurityLevel::SECURITY_BITS` to `SecurityLevel::KAPPA_BITS` (following the paper, it's defined as $\kappa = \log q$ ($q$ is order of the curve), so actually, kappa bits = security bits * 2, so calling it "security bits" is not correct)
    * `SecurityLevel::SECURITY_BYTES` to `SecurityLevel::KAPPA_BYTES`
    * `SecurityLevel::Rid` to `SecurityLevel::KappaBytes` (it's a byte array of $\kappa/8$ bytes; `rid` is a name of variable from the spec, but we use that for other purposes too)
  * Refactor `define_security_level!` macro syntax
    * Old syntax: `define_security_level!(SecurityLevel128 { security_bits = 384 });`
    * New syntax: `define_security_level!(SecurityLevel128 { security_bits: 128 });`
* Add `cggmp24::progress::Stderr`, a tracer that logs all events to stderr, helpful for debugging
* CRT optimization is now enabled by default and cannot be disabled
* Update remaining proofs in `paillier-zk` to accept data by reference to avoid excessive cloning
* Reduce size of `test-data/precomputed_shares.json` by storing aux data separately

TODOs (in following PRs):
- [ ] Refactor NI proofs so they accept NiProof (such that `type NiProof = (Commitment, Proof)`)
- [ ] Remove unused proofs from `paillier-zk`
- [ ] Update `paillier-zk` so the variable names exactly match the spec (e.g. spec says $N_j$, but impl uses a name `key_j`, etc.)
- [ ] Remove unused tests
- [ ] Benchmark and compare revisions
- [ ] Make protocols exactly follow the spec
   Currently, there are many places where sequence of computations is slightly different from the spec, or var naming is different, etc.
- [ ] Update spec so aux gen accepts primes as input (currently spec says to generate primes inside aux gen protocol)
- [ ] Look into changing $\pm \ell$ notation from $\\{- \ell, \dots, +\ell\\}$ to $\\{-(\ell - 1) / 2, \dots, (\ell - 1) / 2\\}$ to match the latest revision of the paper


survived:
@maurges @ThibsG @AurbanDfns I've almost finished cggmp21 -> cggmp24 update, can you please help reviewing this PR? There are a lot of changes, but most essentially we need to thoroughly review that protocol implementation matches the spec, and that the spec matches the paper. You're welcome to take a look at other changes as well (proofread docs, check that API changes make sense, etc.).

The compiled spec may be found [in the comment above](https://github.com/LFDT-Lockness/cggmp21/pull/147#issuecomment-2944912812) (you can also compile it from the sources), the paper [can be found here](https://eprint.iacr.org/2021/060).

In this PR I've updated:
* Aux data generation protocol
  * Code: [cggmp21/src/key_refresh/aux_only.rs](https://github.com/LFDT-Lockness/cggmp21/blob/cggmp24/aux-gen/cggmp21/src/key_refresh/aux_only.rs)
  * Spec: 4.1 Provisioning Protocol (page 25)
  * Paper: Figure 7: Auxiliary Info. & Key Refresh in Three Rounds (page 25)
* Signing (presignature generation + partial signature generation + signature reconstruction):
  * Code: [cggmp21/src/signing.rs](https://github.com/LFDT-Lockness/cggmp21/blob/cggmp24/aux-gen/cggmp21/src/signing.rs)
  * Spec:
    * 4.3 Presigning (page 30)
    * 4.4 Signing (page 33)
  * Paper:
    * Figure 8: ECDSA Threshold Pre-Signing (page 26)
    * Figure 10: ECDSA Signing (page 28)

I suggest that:
* @AurbanDfns reviews that spec matches the paper
* @ThibsG reviews that spec matches the implementation
* @maurges reviews everything :)

Guidelines for implementation <> spec review:
During the review, I suggest that you have little tolerance to any deviations in the implementation from the spec. There are some minor things like order of computations is not exactly the same (I want to change that in my next PR, I just didn't want the diff to be bigger that it is already), if anything else stands out, shoot a comment

Guidelines for paper <> spec review:
As usual, there are a lot of typos in the paper (sign of something is switched, arguments are in wrong order, etc.), so I suggest to be vigilant and check that my understanding of what "was meant by this" and "what authors actually wanted to achieve" is correct

Aurban-dfns:
Few quick remarks on the spec:
- In the paper, ±l is defined as an interval centered at zero, whose size is l (if odd) or l-1 (if even). In the spec, ±l is defined as {−l,...,0,...,l} of size 2l+1. Why this change of convention?
- In generate_Paillier_key, I don't see in Algorithm 1 that the generate primes verify p,q = 3 mod 4 (though it seems to be correctly verified afterward in ProveNI_mod).
- A quick remark on the notations for the fiat-shamir challenge computation. Since it's important for security that the ranges used are also hashed to compute the challenge, why not include L explicitly in the inputs of ChallengeNI?

survived:
> In the paper, ±l is defined as an interval centered at zero, whose size is l (if odd) or l-1 (if even). In the spec, ±l is defined as {−l,...,0,...,l} of size 2l+1. Why this change of convention?

Actually, that a change in notation of the paper. In previous revision ±l was {−l,...,0,...,l}, and only in cggmp24 it's $\pm \ell = \\{-(\ell - 1) / 2, \dots, (\ell - 1) / 2\\}$. I also wonder what was the purpose and if it affects the protocol meaningfully...

UPD: looks like they updated it to match their own implementation of cggmp21, in which they actually don't have negative numbers, this range definition matches more closely to what they have in code. This notation of $\pm \ell$ is used to sample nonces and challenges, I don't think we should have any problems in having twice larger range of $\pm \ell$ then in the paper?

survived:
> In generate_Paillier_key, I don't see in Algorithm 1 that the generate primes verify p,q = 3 mod 4

Algorithm 1 generates a safe prime $p$, such that $p = 2 p' + 1$ where $p'$ is a prime as well. Then $p = 3 \bmod 4$ by definition (if it was the case that $p = 1 \bmod 4$, then $p'$ is even)

survived:
> A quick remark on the notations for the fiat-shamir challenge computation. Since it's important for security that the ranges used are also hashed to compute the challenge, why not include L explicitly in the inputs of ChallengeNI?

@AurbanDfns We do include L in ChallengeNI inputs everywhere where L is present (some proofs don't make use of L, so it's not included here). We provide it a bit differently from other arguments though.

image

Did you mean to say that we should rather provide it as a regular argument, or that we missed `L` somewhere and didn't provide it to ChallengeNI?

spec/main.tex
```diff
@@ -1843,77 +1856,111 @@ \subsection{Provisioning Protocol}
...
+\State Compute Pailier public key $N = pq$
+\State \Return $(N; p, q)$
+\end{algorithmic}
+\end{inlineAlgorithm}
+
+\item
+\begin{inlineAlgorithm}
+\algoName{$\fn{generate\_pedersen\_params}^L() \to (N, s, t; \phi, \lambda)$}
+\algoInputs{security level $L = (\kappa_\rsa, \dots)$}
+\algoOutputs{Pedersen public parameters $N, s, t \in \Z^3$, and secret parameter $\lambda \in \Z$}
```
maurges: Nitpick: phi not mentioned as output. I personally don't remember anymore if it's private or public, but that can be figured out from the further algorithm I guess

cggmp21/src/signing.rs
```diff
@@ -683,13 +720,32 @@ where
...
             K: K_i.clone(),
             G: G_i.clone(),
+            Y: Y_i,
+            A1: A_i1,
+            A2: A_i2,
+            B1: B_i1,
+            B2: B_i2,
         })))
         .await
         .map_err(IoError::send_message)?;
```
maurges: Might instead want to yield between the two multiplications. I don't think this yield there is useful as we're going to yield when sending the messages right after that

cggmp21/src/signing.rs
```diff
@@ -699,25 +755,57 @@ where
         tracer.stage("Prove ψ0_j");
         let R_j = &R[usize::from(j)];
 
-        let psi0 = pi_enc::non_interactive::prove::<D>(
-            &unambiguous::ProofEnc { sid, prover: i },
+        let psi0 = pi_enc_elg::non_interactive::prove::<E, D>(
```
maurges: Might want to rename this to `psi0_ij` for consistency with the paper. The one below to `psi1_ij`. Then maybe in proving call the proofs as `proof_ji`

maurges: Also psi1 doesn't have its stage

cggmp21/src/signing.rs
```diff
@@ -699,25 +755,57 @@ where
...
             },
-            &security_params.pi_enc,
-            &mut *rng,
+            &security_params.pi_enc_elg,
+            rng,
         )
-        .map_err(|e| Bug::PiEnc(BugSource::psi0, e))?;
+        .map_err(|e| Bug::PiEncElg(BugSource::psi0, e))?;
+
+        let psi1 = pi_enc_elg::non_interactive::prove::<E, D>(
```
maurges: Probably worth to add a yield between two proofs

cggmp21/src/signing.rs
```diff
@@ -987,93 +1108,85 @@ where
...
+                pi_elog::Data {
+                    l: &ciphertexts.B1,
+                    m: &ciphertexts.B2,
+                    x: &ciphertexts.Y,
+                    y: &msg.Gamma,
+                    h: &Point::generator().to_point(),
+                },
+                &msg.psi_tilde.0,
+                &msg.psi_tilde.1,
+            )
```
maurges: The stage and the variable still use the name without tilde

cggmp21/src/signing.rs
survived: Yeah I think we only need to add yielding in a long computation where there's no `.await`. If there's one (even if it's sending), it should be okay

spec/main.tex
survived: It's private, knowing phi you can easily find factorization. I added it.

survived:
Force pushing last commit because `git commit -m ...` didn't signoff automatically

cggmp21/src/signing.rs
survived: I'm not sure, there's sending right after computations of these two proofs. It would make sense if we knew that it was the slowest proof, for example, but I didn't do benchmarks yet.

cggmp21/src/signing.rs
survived: since psi0 and psi1 are the same zk proof but with different inputs, I think it makes sense to just let them share the same stage

cggmp21/src/signing.rs
```diff
@@ -1103,49 +1223,42 @@ where
...
-            pi_log::PrivateData {
-                x: &utils::scalar_to_bignumber(&k_i),
-                nonce: &rho_i,
-            },
-            &security_params.pi_log,
-            &mut *rng,
-        )
-        .map_err(|e| Bug::PiLog(BugSource::psi_prime_prime, e))?;
+    tracer.stage("Prove psi_prime");
+    let psi_prime = pi_elog::non_interactive::prove::<E, D>(
```
maurges: Another nitpick about missing `_i` for a proof

cggmp21/src/signing.rs
```diff
@@ -1160,68 +1273,68 @@ where
...
+    let tilde_k_i = SecretScalar::new(&mut (k_i * delta_inv));
+    let tilde_chi_i = SecretScalar::new(&mut (chi_i * delta_inv));
+
+    let Gamma = NonZero::from_point(Gamma).ok_or(Bug::Zero(BugSource::Gamma))?;
+
+    let commitments = round3_msgs
+        .iter_including_me(&my_round3_msg)
+        .map(|m| (delta_inv * m.Delta, delta_inv * m.S))
+        .collect::<Vec<_>>();
+    let commitments = PresignatureCommitments { Gamma, commitments };
```
maurges: Should it be a proper struct instead of a tuple for commitments?

maurges:
It looks like the writeup and the code are consistent

Aurban-dfns:
> Actually, that a change in notation of the paper. In previous revision ±l was {−l,...,0,...,l}, and only in cggmp24 it's ± ℓ = { − ( ℓ − 1 ) / 2 , … , ( ℓ − 1 ) / 2 } . I also wonder what was the purpose and if it affects the protocol meaningfully...

Actually I guess the primary reason is to increase efficiency. 

However, having a larger rage also theoretically means having range proofs that proves weaker statements.
Moreover, the value of ℓ is used to derive the security parameters in Appendix C.1. I don't see any issue right away, but it may be worth digging a bit.

Aurban-dfns:
> Did you mean to say that we should rather provide it as a regular argument, or that we missed `L` somewhere and didn't provide it to ChallengeNI?

I was just saying that having two different notations may be confusing for someone auditing or trying to follow the specs from the paper.

cggmp21/src/signing.rs
survived: I wanted to do that but I couldn't come up with a name for this struct (delta and S are *actual* commtiments, Gamma is just extra data we need). Any suggestions?

survived:
> having a larger rage also theoretically means having range proofs that proves weaker statements

After offline discussion, we came to conclusion that it should not have practical security impacts (e.g. given that security parameters have not been changed between cggmp revisions), however, we do want to improve performance, but the change would imply changing a lot of code (not only cggmp21, but also Paillier encryption, etc.). I will look into this after this PR is merged (added it to TODOs)

survived:
> I was just saying that having two different notations may be confusing for someone auditing or trying to follow the specs from the paper.

Is it two different notations? We just have one argument defined like this for ergonomic reasons, also it's well-defined in the spec

image

I'm leaning to keeping it as is, though if somebody else brings up that it's confusing, we'll change that

cggmp21/src/key_refresh/aux_only.rs
```diff
@@ -387,8 +382,8 @@ where
     for (j, _, d) in decommitments.iter_indexed() {
         tracer.send_msg();
 
-        tracer.stage("Compute П_fac (ф_i^j)");
-        let phi = π_fac::prove::<D>(
+        tracer.stage("Compute П_fac (ψ'_i,j)");
+        let psi_prime = π_fac::prove::<D>(
```
ThibsG, line 386: the spec may be incomplete (Round 3, steps 10 to 12), as it does not refer `π_fac::Aux`, but that could be a misunderstanding from me. Same for `π_fac_security`.

cggmp21/src/signing.rs
maurges: I think calling it a commitment is fine even when there is additional data present

survived: Yes, that's what I did

survived: What about tuples though? Do you suggest any name for a structure?

cggmp21/src/utils.rs
```diff
@@ -215,6 +178,47 @@ pub fn generate_blum_prime(rng: &mut impl rand_core::RngCore, bits_size: u32) ->
...
+        .ok_or(GenPedersenError::BuildCrt)?;
+    let N = (&p * &q).complete();
+    let phi_N = (&p - 1u8).complete() * (&q - 1u8).complete();
+
+    let r = Integer::gen_invertible(&N, rng);
+    let lambda = Integer::random_below(phi_N.clone() >> 2, &mut external_rand(rng));
+
+    let t = r.square().modulo(&N);
+    let s = crt
+        .exp(&t, &crt.prepare_exponent(&lambda))
```
ThibsG, line 200: `s = t.exp(λ) mod N` is done by `.exp(&t, &crt.prepare_exponent(&lambda))`.

but `prepare_exponent(..)` is using `phi_n` and not `N` as mentioned in 4.1.1.

cggmp21/src/signing.rs
maurges: Options:
1. PresignatureCommitment and PresignatureCommitmentPack - like in frost
2. PresignatureCommitment and PresignatureCommitments - might be confusing
3. PresignatureCommitment and PresignaturePublicData - now Gamma is less awkward I guess

I think I like option 3

ThibsG (commented):
Partial review for provisioning protocol

cggmp21/src/key_refresh/aux_only.rs
survived: `π_fac::Aux` (2nd argument to `π_fac::prove`) is actually $R_j$ in the paper (2nd argument in $\mathtt{ProveNI}_\mathtt{fac}$ function from the spec)

survived: and `π_fac_security` is $L$ from the spec (notice $L$ in $\mathtt{ProveNI}_\mathtt{fac}^L$)

cggmp21/src/utils.rs
survived: Right, in the implementation we use CRT optimization that is not described in the spec, I will update it

survived: Added a comment in the spec (best way I found to mention it)
image

cggmp21/src/key_refresh/aux_only.rs
ThibsG: thanks for the precision!

cggmp21/src/utils.rs
ThibsG: (for some reason I cannot close the thread of comments sorry ^^)

cggmp21/src/signing.rs
```diff
@@ -911,8 +1061,8 @@ where
             },
             &R_j.into(),
             pi_aff::Data {
-                key0: &enc_j,
-                key1: &dec_i,
+                key_j: enc_j,
+                key_i: &dec_i,
```
ThibsG, line 1073: Reading the paper p31 (round 2, step 14 and 15), I assume `key_j` and `key_i` are respectively corresponding to `N_j` and `N_i`, right? If so, why not using directly the variable `N_j`, is there some perf reason here or maybe they are not the same at all?

cggmp21/src/signing.rs
```diff
@@ -987,93 +1114,86 @@ where
...
+            let psi_j_invalid = pi_aff::non_interactive::verify::<E, D>(
+                &unambiguous::ProofPsi {
+                    sid,
+                    prover: j,
+                    hat: false,
+                },
+                &R_i.into(),
+                pi_aff::Data {
+                    key_j: &dec_i,
+                    key_i: enc_j,
```
ThibsG, line 1160: I'm unsure about these parameters because above they were inverted.

Just underline this for you to double check if there is an inversion or if it's totally fine.

cggmp21/src/signing.rs
```diff
@@ -987,93 +1114,86 @@ where
...
+                None
+            }
+        });
     if !faulty_parties.is_empty() {
         return Err(SigningAborted::InvalidPsi(faulty_parties).into());
     }
+    runtime.yield_now().await;
 
     // Step 2
     tracer.stage("Compute Gamma, Delta_i, delta_i, chi_i");
```
ThibsG, line 1207: right below there is this line:
`let Gamma = Gamma_i + round2_msgs.iter().map(|msg| msg.Gamma).sum::<Point<E>>();`

In the spec there is no such addition. Maybe it's normal because `round2_msgs` is built excluding `i` index?

cggmp21/src/signing.rs
```diff
@@ -1160,68 +1273,68 @@ where
...
     if !faulty_parties.is_empty() {
         return Err(SigningAborted::InvalidPsiPrimePrime(faulty_parties).into());
     }
+    runtime.yield_now().await;
 
     // Step 2
     tracer.stage("Calculate presignature");
     let delta = delta_i + round3_msgs.iter().map(|m| m.delta).sum::<Scalar<E>>();
     let Delta = Delta_i + round3_msgs.iter().map(|m| m.Delta).sum::<Point<E>>();
+    let S = S_i + round3_msgs.iter().map(|m| m.S).sum::<Point<E>>();
```
ThibsG, line 1313: I guess it's the same here as above. The addition is not in the spec, but the sum is incomplete if we don't add `S_i`, right?

ThibsG (commented):
Presigning n-out-of-n review

cggmp21/src/signing.rs
survived: Yes, exactly, key_j and key_i are $N_j$ and $N_i$ respectively. Sometimes the naming is not exactly as in the paper, but I think it's a good idea to update it here

survived: Actually, I started updating variable names to match exactly the spec, and turns out that variable naming don't follow the spec very often and the resulting diff will be huge. I will update car names in another PR, added that to TODO

cggmp21/src/signing.rs
survived: I also like the 3rd option, thanks!

cggmp21/src/signing.rs
survived: Actually, it's structure fields, the order in which they are specified doesn't matter, you can swap those lines, nothing will change

survived: I have a task in my TODOs to make code follow exactly the spec though, I didn't want to do that in this PR to avoid even larger diff

cggmp21/src/signing.rs
survived: Yes, exactly `round2_msgs` only includes messages received from other parties, so we need to add our Gamma_i

cggmp21/src/signing.rs
survived: Yep, exactly, message store only has messages received from other parties

cggmp21/src/signing.rs
ThibsG: > Actually, it's structure fields, the order in which they are specified doesn't matter, you can swap those lines, nothing will change

It's not about the order but the mapping.
Here `key_j` -> `dec_i` while previously (line 1064) `key_j` -> `enc_j`

survived: Ah, yes, this is expected and reflects the spec:

The zk proof definition:
image

image

Then when we prove in 2nd round:
image

And when we verify in the 3rd round:
image

So you can see that the zk proof (both prove and verify) accept $N_j, N_i$ (in this specific order), in round 2 on proving stage we provide $N_j, N_i$, but in round 3 on verifying stage they get flipped and we provide $N_i, N_j$

ThibsG: All good then 😉

maurges (approved):

survived:
Thanks everyone!

PR #149, opened 2025-08-06
Fix/ell

jfdreis:
## Updates to the ±ℓ Notation

The following changes were made to update the **±ℓ notation**:

- **Previous notation meant:**  `{ -l, ..., l }`

- **New notation means:**  `{ -l/2, ..., l/2 }` (when l is even) , `{ -(l-1)/2, ..., (l-1)/2 }` (when l is odd)

### 1. `common.rs`

- **Fixed a typo** in `Integer::from_rng_pm`:  
  Previously, it produced values in the range `[-range, range)`, instead of `[-range, range]` as expected by the **cggmp21** version.

- **Added new traits:**
  - `Integer::from_rng_half_pm` – Generates values according to the new notation (**cggmp24**).
  - `Integer::is_in_half_pm` – Checks if a value lies within the new range.

- **Added tests** to verify that these functions work as expected.

### 2. Updated Files

The following files now use `Integer::from_rng_half_pm` and `Integer::is_in_half_pm` instead of `Integer::from_rng_pm` and `Integer::is_in_pm`:

- `signing.rs`
- `no_small_factors.rs`
- `paillier_encryption_in_range.rs`
- `paillier_encryption_in_range_with_el_gamal.rs`
- `paillier_affine_operation_in_range.rs`

### 3. Observations

- **`paillier_encryption_in_range.rs`**  
  Although this ZKP is not used in **cggmp24**, it was kept because it appears in the protocol.

- **`group_element_vs_paillier_encryption_in_range.rs`**  
  This file was **not updated**, as it is not used in **cggmp24** (the latest version of the paper), so it retains the previous notation.


paillier-zk/src/common.rs
```diff
@@ -224,14 +234,42 @@ impl IntegerExt for Integer {
     fn from_rng_pm<R: rand_core::RngCore>(range: &Self, rng: &mut R) -> Self {
         let mut rng = fast_paillier::utils::external_rand(rng);
         let range_twice = range.clone() << 1u32;
-        range_twice.random_below(&mut rng) - range
+        let range_twice_plus_one = range_twice + Integer::ONE.clone();
```
survived: You probably don't need `clone` here given that lhs is owned

survived: Ahh that was a bug, thanks for fixing it! We should have tests that check that we are able to randomly sample all possible integers on small ranges to make sure that never happens again.

paillier-zk/src/common.rs
```diff
@@ -224,14 +234,42 @@ impl IntegerExt for Integer {
...
         let minus_range = -range.clone();
         minus_range <= *self && self <= range
     }
 
+    fn from_rng_half_pm<R: rand_core::RngCore>(range: &Self, rng: &mut R) -> Self {
+        let mut rng = fast_paillier::utils::external_rand(rng);
+
+        if range.clone().is_even() {
+            let half_range = range.clone() >> 1u32;
+            let range_plus_one = range.clone() + Integer::ONE.clone();
```
survived: also a redundant `clone` given that lhs is owned

paillier-zk/src/common.rs
survived: You already added it below! Thanks, it's perfect! 🔥

survived:
```suggestion
        let range_twice_plus_one = range_twice + Integer::ONE;
```

survived:
> `group_element_vs_paillier_encryption_in_range.rs` \
This file was not updated, as it is not used in cggmp24 (the latest version of the paper), so it retains the previous notation.

We should delete the proof before we forget to do that again. Let me do that, then you can merge `cggmp24/m` into your branch, it should not introduce any conflicts.

UPD: you can merge `cggmp24/m` into your branch and you won't have this proof anymore

paillier-zk/src/common.rs
```diff
@@ -224,14 +234,42 @@ impl IntegerExt for Integer {
...
         let range_twice = range.clone() << 1u32;
-        range_twice.random_below(&mut rng) - range
+        let range_twice_plus_one = range_twice + Integer::ONE.clone();
+        range_twice_plus_one.random_below(&mut rng) - range
     }
 
     fn is_in_pm(&self, range: &Self) -> bool {
         let minus_range = -range.clone();
         minus_range <= *self && self <= range
     }
```
survived: Once unused ZK proof is removed, we can remove those methods, as they are not used anywhere else

paillier-zk/src/common.rs
```diff
@@ -224,14 +234,42 @@ impl IntegerExt for Integer {
...
+            return lower_bound <= *self && *self <= upper_bound;
+        }
+
+        let range_minus_one = range.clone() - Integer::ONE;
+        let upper_bound = range_minus_one >> 1u32;
+        let lower_bound = -upper_bound.clone();
+        lower_bound <= *self && *self <= upper_bound
+    }
+
     fn signed_modulo(&self, n: &Self) -> Self {
```
survived: I think we need to remove this method. It uses old notation, so it's confusing to keep it. It's not used anywhere anyways.

survived:
@jfdreis I think we also need to update `cggmp21::utils::scalar_to_bignumber` which is used to convert a scalar to bignumber:

https://github.com/LFDT-Lockness/cggmp21/blob/74795ea0d6339205a8209a857a881213469e6f86/cggmp21/src/utils.rs#L14-L17

The problem that resulting bignumber is in $[0; q)$, where as in the protocol it needs to be in $\pm q = \\{-(q-1)/2, \dots, (q-1)/2\\}$.

E.g. in proof $\Pi_\text{aff}$ private inputs, we provide `x: &utils::scalar_to_bignumber(&gamma_i)`:

https://github.com/LFDT-Lockness/cggmp21/blob/74795ea0d6339205a8209a857a881213469e6f86/cggmp21/src/signing.rs#L1051-L1056

However, protocol requires that $x \in \pm 2^\ell$ where $\ell$ is set to $|q|$:

https://github.com/LFDT-Lockness/cggmp21/blob/74795ea0d6339205a8209a857a881213469e6f86/spec/main.tex#L605-L606

We must be breaking protocol by providing $x \in [0; q)$.

So I think we should rename this function to `cggmp21::utils::scalar_to_pm_bignumber`. We should first convert scalar absolute value to big number `x` and then return `x` if `x <= (q - 1) / 2` (or `2*x < q`), otherwise return `x - q`.

paillier-zk/src/common.rs
jfdreis: agreed. I will remove those two methods.

paillier-zk/src/common.rs
jfdreis: signed_modulo is used in some doc tests in paillier_affine_operation_in_range.rs, but I think we can remove.

jfdreis:
@survived I agree as the plaintexts are always in those kind of ranges around zero.
I think that the same reasoning should apply for the CGGMP21 version since the following is stated there:  *“We represent integers modulo N in the interval {−N/2, …, N/2} (rather than the canonical representation)”*.

I am making the changes for the current version CGGMP24

survived:
> the same reasoning should apply for the CGGMP21 version since the following is stated there: _“We represent integers modulo N in the interval {−N/2, …, N/2} (rather than the canonical representation)”_.

You're right, it does. That's a bug of implementation that we converted scalars to bignumbers in this way.

survived (dismissed):
I think everything is good. @maurges do you want to take a look?

paillier-zk/src/common.rs
```diff
@@ -221,15 +225,31 @@ impl IntegerExt for Integer {
         i + 1
     }
 
-    fn from_rng_pm<R: rand_core::RngCore>(range: &Self, rng: &mut R) -> Self {
+    fn from_rng_half_pm<R: rand_core::RngCore>(range: &Self, rng: &mut R) -> Self {
         let mut rng = fast_paillier::utils::external_rand(rng);
-        let range_twice = range.clone() << 1u32;
-        range_twice.random_below(&mut rng) - range
+
+        if range.clone().is_even() {
```
maurges: Unnecessary clone

paillier-zk/src/common.rs
```diff
@@ -221,15 +225,31 @@ impl IntegerExt for Integer {
...
+        let range_minus_one = range.clone() - Integer::ONE;
+        let half_range_minus_one = range_minus_one >> 1u32;
+        range.clone().random_below(&mut rng) - half_range_minus_one
     }
 
-    fn is_in_pm(&self, range: &Self) -> bool {
-        let minus_range = -range.clone();
-        minus_range <= *self && self <= range
+    fn is_in_half_pm(&self, range: &Self) -> bool {
+        if range.clone().is_even() {
```
maurges: Unnecessary clone

maurges (changes requested):
A couple of nitpicks, but looks good otherwise

jfdreis:
I noticed that before I forgot to delete [fn signed_modulo](https://github.com/LFDT-Lockness/cggmp21/blob/9ddcc57417206178b8ee07398db158a66b37b712/paillier-zk/src/common.rs#L184C3-L184C47). 
It is now deleted.

maurges:
Please avoid force-pushing, this makes it very hard to see new changes since we did a review

survived (approved):

maurges (approved):

maurges (approved):

survived:
Thanks for your contribution @jfdreis! It's very helpful!

PR #150, opened 2025-08-12
Remove unused ZK proof

survived:


survived:
I suggest that we ignore clippy errors to avoid merge conflicts, they are going to be fixed in #149

maurges (approved):

Issue #151, opened 2025-08-12
CGGMP24 TODOs


survived:
We have finished initial upgrade to CGGMP24 implementation, but we have left something to do before proper release:

- [x] Remove unused proofs from `paillier-zk` \
  #150 
- [x] Changing $\pm \ell$ notation from $\\{- \ell, \dots, +\ell\\}$ to $\\{-(\ell - 1) / 2, \dots, (\ell - 1) / 2\\}$ to match the latest revision of the paper \
  #149 
- [x] Raname crates, folders, code, i.e. `s/cggmp21/cggmp24/g` \
   #152
- [x] Update README
   It got outdated, e.g. it mentions 5+1 signing protocol and key refresh (which was removed) \
   #153
- [ ] Change repo name to `cggmp24`
- [ ] Make sure all links work \
    E.g. link to the spec in readme is currently broken
- [ ] Remove security parameter `Q` from the spec that is used as domain space of challenges \
   Paper now samples challenges from $\pm q$
- [x] Refactor NI proofs so they accept NiProof (such that `type NiProof = (Commitment, Proof)`) \
   #154 \
   #155 \
   #156 \
   #157 \
   #158 \
   #160
- [ ] Update `paillier-zk` so the variable names exactly match the spec (e.g. spec says $N_j$, but impl uses a name `key_j`, etc.)
- [ ] Benchmark and compare revisions
- [ ] Make protocols exactly follow the spec
   Currently, there are many places where sequence of computations is slightly different from the spec, or var naming is different, etc.
- [ ] Update spec so aux gen accepts primes as input (currently spec says to generate primes inside aux gen protocol)

Divyateja2709:
Hey @survived  
I'd like to take on the 'Update paillier-zk variable names to match the spec' task from this list. I've been reading the CGGMP24 paper and the spec alongside the implementation. I'll go through each ZK proof file, map every variable name to the spec notation, and open a PR with the renames. Should I treat the PDF spec at lfdt-lockness.github.io as the source of truth for naming?

PR #153, opened 2025-08-25
Update README.md

jfdreis:
### Update README.md for CGGMP24

This update addresses one of the TODOs in the Issue [CGGMP24 TODOs](https://github.com/LFDT-Lockness/cggmp21/issues/151#issue-3313573246).

### Changes

- Removed mentions to the **5 + 1 signing protocol**.
- Update the mentions to **key refresh** : 
   - remove it from [this crate implements](https://github.com/LFDT-Lockness/cggmp21/blob/8cd2b4da5759a6fea90fbb7698715fd779a8e9eb/README.md?plain=1#L37)
   - added it under [This crate **does not** (currently) support](https://github.com/LFDT-Lockness/cggmp21/blob/8cd2b4da5759a6fea90fbb7698715fd779a8e9eb/README.md?plain=1#L47)
   - Note that CGGMP24 still supports key refresh.
- As [CGGMP24](https://eprint.iacr.org/2021/060.pdf) now **supports threshold (t-out-of-n) keys** (see Section F.1), removed all statements claiming that it was not supported in CGGMP24

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,6 +48,11 @@ This crate **does not** (currently) support:
 
 Our implementation has been audited by Kudelski. Report can be found [here][report].
 
+> About notion of threshold and non-threshold keys: originally, the CGGMP24 paper does not specify
+protocols for threshold t-out-of-n keys with arbitrary `t`, and only works with non-threshold
+n-out-of-n keys. We have added support for arbitrary threshold $2 \le t \le n$, however, we made
+it possible to opt out thresholdness so original CGGMP24 protocol can be carried out if needed.
+
 ## Running the protocol
 
 ### Networking
@@ -228,7 +233,13 @@ However, you may opt for them by enabling `spof` feature, then you can use `trus
 for key import and `key_share::reconstruct_secret_key` for key export.
 
 ## Differences between the implementation and CGGMP24
-There are small differences in the implementation compared to the original paper [CGGMP24] (mostly typo fixes);
+[CGGMP24] only defines a non-threshold protocol. To support general thresholds,
+we defined our own CGGMP24-like key generation and threshold signing
+protocols. However, we keep both
+threshold and non-threshold versions of the protocols in the crate, so if you opt for the non-threshold
+protocol, you will be running the original protocol defined in the paper.
+
+There are other (small) differences in the implementation compared to the original paper (mostly typo fixes);
 they are all documented in [the spec].
 
 [CGGMP24]: https://ia.cr/2021/060
--- a/README.md
+++ b/README.md
@@ -26,15 +26,13 @@
 
 <!-- TOC ENDS -->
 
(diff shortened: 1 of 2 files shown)


survived:
> As [CGGMP24](https://eprint.iacr.org/2021/060.pdf) now supports threshold (t-out-of-n) keys (see Section F.1), removed all statements claiming that it was not supported in CGGMP24

I think what we had in readme originally was not phrased right. I didn't mean to say "cggmp21 is incompatible with threshold setting" (it always has been), what I meant is that cggmp paper does not specify threshold protocol, it only defines n-out-of-n protocols. Therefore, we had to do significant changes in the protocol (compared to what defined in the paper) to support threshold keygen/signing.

All F.1 says is "Our protocol can be extended to the t-out-of-n threshold case", and very briefly explains how it can be done. So cggmp paper still doesn't provide threshold protocol, and if one needs it, they have to introduce changes into the protocol.

README.md
```diff
@@ -45,17 +43,11 @@ We also provide auxiliary tools like:
...
 * Identifiable abort
-* The (5+1)-round signing protocol
 
 Our implementation has been audited by Kudelski. Report can be found [here][report].
 
-> About notion of threshold and non-threshold keys: originally, CGGMP24 paper does not have support of
-arbitrary `t` and only works with non-threshold n-out-of-n keys. We have added support of arbitrary
-threshold $2 \le t \le n$, however, we made it possible to opt out thresholdness so original CGGMP24
-protocol can be carried out if needed.
-
```
survived: This is still mostly true, though phrasing can be improved:

- cggmp doesn't specify threshold protocols
- we have added support of arbitrary threshold
- it's still possible to opt out thresholdness so original cggmp24 protocol is carried out

README.md
```diff
@@ -236,13 +228,7 @@ However, you may opt for them by enabling `spof` feature, then you can use `trus
...
 
 ## Differences between the implementation and CGGMP24
-[CGGMP24] only defines a non-threshold protocol. To support general thresholds,
-we defined our own CGGMP24-like key generation and threshold signing
-protocols. However, we keep both
-threshold and non-threshold versions of the protocols in the crate, so if you opt for the non-threshold
-protocol, you will be running the original protocol defined in the paper.
-
-There are other (small) differences in the implementation compared to the original paper (mostly typo fixes);
+There are small differences in the implementation compared to the original paper [CGGMP24] (mostly typo fixes);
```
survived: This is still true as well:

- cggmp only defines a non-threshold protocol (and briefly says that adding thresholdness is possible)
- library is structured in the way that user can opt out thresholdness, so original protocol is carried out

README.md
jfdreis: I will revert the changes and rephrase it slightly.

survived (approved):
Perfect, thanks! 🙏

PR #154, opened 2025-08-26
Refactor Non-Interactive Proofs to use type NiProof

jfdreis:
This PR partially addresses one of the TODOs in [CGGMP24 TODOs](https://github.com/LFDT-Lockness/cggmp21/issues/151#issue-3313573246):

Refactoring Non Interactive Proofs such that they accept `type NiProof = (Commitment, Proof)`

Currently, this change has been applied to the Non-Interactive ZK-proof of Paillier-Blum modulus.

### DIFF
--- a/cggmp24/src/key_refresh/aux_only.rs
+++ b/cggmp24/src/key_refresh/aux_only.rs
@@ -93,7 +93,7 @@ pub struct MsgRound2<L: SecurityLevel> {
 pub struct MsgRound3 {
     /// $\psi_i$
     // this should be L::M instead, but no rustc support yet
-    pub mod_proof: π_mod::non_interactive::Proof<{ crate::security_level::M }>,
+    pub mod_proof: π_mod::NiProof<{ crate::security_level::M }>,
     /// $\phi_i^j$
     pub fac_proof: π_fac::Proof,
 }
--- a/cggmp24/src/key_refresh/aux_only.rs
+++ b/cggmp24/src/key_refresh/aux_only.rs
@@ -2,7 +2,7 @@ use digest::Digest;
 use futures::SinkExt;
 use paillier_zk::{
     no_small_factor::non_interactive as π_fac,
-    paillier_blum_modulus::{self as π_mod},
+    paillier_blum_modulus as π_mod,
     rug::{Complete, Integer},
 };
 use rand_core::{CryptoRng, RngCore};
@@ -93,7 +93,7 @@ pub struct MsgRound2<L: SecurityLevel> {
 pub struct MsgRound3 {
     /// $\psi_i$
     // this should be L::M instead, but no rustc support yet
-    pub mod_proof: π_mod::NiProof<{ crate::security_level::M }>,
+    pub mod_proof: π_mod::non_interactive::Proof<{ crate::security_level::M }>,
     /// $\phi_i^j$
     pub fac_proof: π_fac::Proof,
 }
@@ -435,7 +435,6 @@ where
         &decommitments,
         &shares_msg_b,
         |j, decommitment, proof_msg| {
-            let niproof = &proof_msg.mod_proof;
             π_mod::non_interactive::verify::<{ crate::security_level::M }, D>(
                 &unambiguous::ProofMod {
                     sid,
@@ -443,7 +442,7 @@ where
                     prover: j,
                 },
                 π_mod::Data { n: &decommitment.N },
-                niproof,
+                &proof_msg.mod_proof,
             )
             .is_err()
         },
--- a/cggmp24/src/key_refresh/aux_only.rs
+++ b/cggmp24/src/key_refresh/aux_only.rs
@@ -2,7 +2,7 @@ use digest::Digest;
 use futures::SinkExt;
 use paillier_zk::{
     no_small_factor::non_interactive as π_fac,
-    paillier_blum_modulus as π_mod,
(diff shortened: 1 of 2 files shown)


paillier-zk/src/paillier_blum_modulus.rs
```diff
@@ -115,6 +114,14 @@ pub struct Proof<const M: usize> {
     pub points: [ProofPoint; M],
 }
 
+/// The Non-interactive ZK proof. Combines commitment and proof
+#[derive(Debug, Clone)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+pub struct NiProof<const M: usize> {
```
maurges, line 121: In other zk-proofs we call this struct simply `Proof`. Interactive and non-interactive are distinguished by the module name

The same can be said about calling the variable `niproof` instead of just `proof`

cggmp24/src/key_refresh/aux_only.rs
```diff
@@ -438,16 +435,15 @@ where
         &decommitments,
         &shares_msg_b,
         |j, decommitment, proof_msg| {
-            let (comm, proof) = &proof_msg.mod_proof;
+            let niproof = &proof_msg.mod_proof;
```
maurges: Useless local?

cggmp24/src/key_refresh/aux_only.rs
```diff
@@ -2,7 +2,7 @@ use digest::Digest;
 use futures::SinkExt;
 use paillier_zk::{
     no_small_factor::non_interactive as π_fac,
-    paillier_blum_modulus as π_mod,
+    paillier_blum_modulus::{self as π_mod},
```
maurges: I don't see why this change was made

paillier-zk/src/paillier_blum_modulus.rs
```diff
@@ -252,20 +259,21 @@ pub mod non_interactive {
...
         shared_state: &impl udigest::Digestable,
         data: Data,
-        commitment: &Commitment,
-        proof: &Proof<M>,
+        niproof: &NiProof<M>,
     ) -> Result<(), InvalidProof> {
+        let commitment = &niproof.commitment;
+        let proof = &niproof.proof;
         let challenge = challenge::<M, D>(shared_state, data, commitment);
         super::interactive::verify(data, commitment, &challenge, proof)
```
maurges: Unnecessary locals, in other modules we directly pass the struct fields to the subfunctions

maurges:
Thanks for the patch, I left a couple of notes

cggmp24/src/key_refresh/aux_only.rs
jfdreis: At some point, when I was making changes I had
`    paillier_blum_modulus::{self as π_mod, NiProof},
`

When I deleted NiProof I forgot to undo that change.

paillier-zk/src/paillier_blum_modulus.rs
jfdreis: Thanks! I’ll rename the struct to `Proof` inside the non_interactive module, and update variables from `niproof` to just `proof`.

survived: Actually, what I meant is that we have `NiProof` indeed, but placed in parent module.

Currently, there's already `paillier_blum_modulus::Proof` defined: 

https://github.com/LFDT-Lockness/cggmp21/blob/49e7139cd0bd14b3b789104ef3aecf1713465930/paillier-zk/src/paillier_blum_modulus.rs#L105-L116

What I suggest is that next to it we have:

```rust
type NiProof<const M: usize> = (Commitment, Proof<M>);
```

and then that `non_interactive` module uses `NiProof` in proving/verification

survived: I don't think it's great to have `paillier_blum_modulus::Proof` and `paillier_blum_modulus::non_interactive::Proof`. In this case, we need to move `paillier_blum_modulus::Proof` to `paillier_blum_modulus::interactive::Proof` (along with commitments and all other types) and only then it's seems okay to keep `paillier_blum_modulus::non_interactive::Proof`.

survived: but I'm satisfied with how types are organized right now, so I'd prefer that we just defined `NiProof` in parent module

maurges (dismissed):

paillier-zk/src/paillier_blum_modulus.rs
maurges: I like your suggestion, we could do it. But I'm also content with how it's organized now

jfdreis: I can revert some of the changes, if you prefer. Namely, I can move back the `NiProof` struct to the parent module and make the necessary changes for coherence.

I will keep the variables names as `proof` and not `niproof`.

Also, once a new design is settled I would replicate it in other zkproofs, if that is ok for you.

survived: Yes, please, I'd prefer that to be done this way, I find it more consistent as an API organization.

survived (approved):
Looks great, thank you!

Just a little remark for future PRs: I'd say that it's preferable to write `ni_proof` instead of `niproof`, it'd be more correct translation from camel case `NiProof`

PR #155, opened 2025-08-28
Add NiProof struct in zk dlog with el gamal commit

jfdreis:
Similar to PR [#154](https://github.com/LFDT-Lockness/cggmp21/pull/154#issue-3356180215), this PR partially addresses one of the TODOs from [CGGMP24TODOs](https://github.com/LFDT-Lockness/cggmp21/issues/151#issue-3313573246): Refactor NI proofs so they accept NiProof `struct NiProof = (Commitment, Proof)`.

This change was applied to the ZK-proof of discrete log with El-Gamal commitment.

I still call `proof` for some variables whose type is `NiProof`. If you prefer, I can change those variables name to `ni_proof`.

### DIFF
--- a/cggmp24/src/signing.rs
+++ b/cggmp24/src/signing.rs
@@ -216,7 +216,7 @@ pub mod msg {
         /// $\hat F_{j,i}$
         pub hat_F: fast_paillier::Ciphertext,
         /// $\psi_i$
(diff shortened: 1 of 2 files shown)


survived (approved):
Looks great!

survived:
> I still call proof for some variables whose type is NiProof

I'd actually prefer to keep them called `proof`, I think whether it's NI or not is immediately evident from the context.

PR #156, opened 2025-08-29
Add NiProof struct in zk enc: paillier encryption in range

jfdreis:
Similar to PR #154, #155, this PR partially addresses one of the TODOs from #151 : Refactor NI proofs so they accept NiProof.

This change was applied to the ZK-proof of Paillier encryption in range


survived (approved):
Thanks!

PR #157, opened 2025-08-29
Add NiProof struct in zk enc-elg: paillier enc in range with el-gamal commit

jfdreis:
Similar to PR #154, #155 , #156 this PR partially addresses one of the TODOs from #151  : Refactor NI proofs so they accept NiProof.

This change was applied to the ZK-proof of Paillier encryption in range with el-gamal commit


survived (approved):
It's aligned with previous PRs, thanks for submission!

PR #158, opened 2025-08-29
Add NiProof struct in zk affine operation with group commit in range

jfdreis:
Similar to PR #154, #155 , #156, #157 this PR partially addresses one of the TODOs from #151  : Refactor NI proofs so they accept NiProof.

This change was applied to the ZK-proof of Paillier operation with group commitment in range


survived (approved):
Changes are aligned with previous PRs. Thanks!

PR #160, opened 2025-09-01
Move NiProof struct to parent module in zk factoring of a RSA modulus

jfdreis:
This PR partially addresses one of the TODOs from #151  : Refactor NI proofs so they accept NiProof.

The `Proof` struct for the Non_interactive ZK-proof for factoring of a RSA modulus was already define under the `non_interactive` module. I moved it to the parent module, rename it to `NiProof`, and update the code accordingly.

Together with PRs #154, #155, #156 , #157, and #158, this completes the Refactor of NI Proofs. (see #151).


survived (approved):
Perfect! Thanks for updating all the proofs!

PR #169, opened 2025-11-24
Switch to new fast-paillier with backend choice

maurges:
- Add the same backend choice here
- Modify CI to test with both backends
- Remove some util functions as they now exist on the backend integer itself

Moved from the private security fix fork.

The unreviewed commit is 250812a6e430d3d6b08976a87d61325ea3ea44c3


maurges:
Closed to rebase on master

PR #172, opened 2026-02-16
Add serde traits for presignature public data and commitment

alfonzii:
To be able to serialize presignatures as a whole also in CGGMP24, we need to derive serde traits for public data.

### DIFF
--- a/cggmp24/src/signing.rs
+++ b/cggmp24/src/signing.rs
@@ -192,7 +192,7 @@ pub struct Presignature<E: Curve> {
 /// Public part of the presignature that can be used to verify partial signatures from other parties
 ///
 /// They are used to validate partial signature produced by the signers from a presignature
-#[derive(Clone, Debug, PartialEq, Eq)]
+#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
 pub struct PresignaturePublicData<E: Curve> {
     /// $\Gamma$ presignature commitment
     pub Gamma: NonZero<Point<E>>,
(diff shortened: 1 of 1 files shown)


cggmp24/src/signing.rs
```diff
@@ -192,7 +192,7 @@ pub struct Presignature<E: Curve> {
 /// Public part of the presignature that can be used to verify partial signatures from other parties
 ///
 /// They are used to validate partial signature produced by the signers from a presignature
-#[derive(Clone, Debug, PartialEq, Eq)]
+#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
 pub struct PresignaturePublicData<E: Curve> {
```
survived, line 196: This alone is not sufficient, you also need to add this attribute, otherwise derived trait impl will be unusable

```suggestion
#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
#[serde(bound = "")]
pub struct PresignaturePublicData<E: Curve> {
```

cggmp24/src/signing.rs
```diff
@@ -201,7 +201,7 @@ pub struct PresignaturePublicData<E: Curve> {
 }
 
 /// Presignature commitment, used to verify partial signature correctness
-#[derive(Clone, Debug, PartialEq, Eq)]
+#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
```
survived, line 204: here as well

survived (commented):
Thanks for submitting it! Due to org policies, we can only merge PR if:
- commits are signed-off (there's a [failing workflow](https://github.com/LFDT-Lockness/cggmp21/pull/172/checks?check_run_id=63766730800) with instructions how to fix it)
- commits are [GPG signed](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits)

Could you sign-off commits and GPG-sign them?

Also, if you want this change to be released to crates-io after this PR is merged, could you update the lib version to `0.7.0-alpha.4`?

PR #173, opened 2026-03-16
Add support of secp384r1 curve

survived:
* Added secp384r1 curve from https://github.com/LFDT-Lockness/generic-ec/pull/59 to the cggmp.
* There's no HD derivation algorithm for this curve, so I had to restructure tests which were expecting each curve to be HD-compatible.
* Also added 192 bits security level matching the added curve
* Updated all protocols to work with any digest (there was a requirement that output of digest had to be 32 bytes)
* Added 3840 bits primes to the precomputed primes, matching the new security level
* Added secp384r1 key shares to pregenerated key shares 
* <s>Size of pregenerated shares is over 400 MB, I had to move it to LFS</s>
* Optimized size of pregenerated shares file

I noticed that it's very easy to make a mistake and provide wrong security level/digest to secp384r1 curve. Previously, we were always using the same security level/digest pretty much everywhere, but now we have different curves that should be used with different digest/security level.

In next PR, I want to add `trait Suite { type Digest; type Curve; type SecLevel; }` and then change api so it takes a cryptosuite instead of curve, e.g. 
* `cggmp24::keygen::<cggmp24::suites::Secp256k1>(eid, i, n)` 
* instead of `cggmp24::keygen::<cggmp24::supported_curves::Secp256k1>(eid, i, n)`


survived:
@maurges @nikita-dfns can you look at changes I pushed in 89397b8b4999aa4b902f1a6a01a2801eb130317c before I update all tests to work with HD in similar way?

tests/src/lib.rs
```diff
@@ -94,47 +93,45 @@ where
...
+    pub static ref CACHED_SHARES: PrecomputedKeyShares = {
+        // note: serialized pregenerated shares take so much space, my (virtualized) compiler gets killed
+        // on RAM overuse when trying to `include_str!` the file into the binary.
+        let mut path = std::path::PathBuf::from(env!("CARGO_MANIFEST_DIR"));
+        path.push("../test-data/precomputed_shares.json");
+
+        let file = std::fs::File::open(path).unwrap();
+        let reader = std::io::BufReader::new(file);
+        serde_json::from_reader(reader).unwrap()
+    };
```
maurges: Actually since we already have this mechanism, I would prefer other pregenerated data to be loaded at runtime as well. This should improve link time by at least a bit

Cargo.toml
```diff
@@ -54,3 +54,16 @@ num-bigint.opt-level = 3
...
+
+[patch.crates-io.generic-ec-curves]
+git = "https://github.com/sridhar-panigrahi/generic-ec.git"
+rev = "5168020e97af1148f5703ad5c5b8b35c98511a8d"
+[patch.crates-io.generic-ec]
+git = "https://github.com/sridhar-panigrahi/generic-ec.git"
+rev = "5168020e97af1148f5703ad5c5b8b35c98511a8d"
+[patch.crates-io.generic-ec-zkp]
+git = "https://github.com/sridhar-panigrahi/generic-ec.git"
+rev = "5168020e97af1148f5703ad5c5b8b35c98511a8d"
```
maurges: Can already be replaced with the lockness repo. And maybe add a todo so that we don't forget to remove those after release? We are going to wait for generic-ec release before merging this, right?

maurges:
Looks good. I actually like the test changes

Cargo.toml
survived: I usually leave a TODO, forgot this time. Yes, generic-ec is already released, `hd-wallet` waits for your review in https://github.com/LFDT-Lockness/hd-wallet/pull/19

survived:
@maurges pushed suggested changes + reworked the precomputed shares storage. Now it's optimal, we do not store the duplicate information, as result size of file dropped from 401M to 116K. As result, there's more code for key shares reconstruction (as they are stored in disassembled state), but I think it's worth it.

tests/src/lib.rs
```diff
@@ -93,131 +88,284 @@ where
...
-            include_str!("../../test-data/precomputed_shares.json")
-        ).unwrap();
-    pub static ref CACHED_PRIMES: PregeneratedPrimes =
-        PregeneratedPrimes::from_serialized(
-            include_str!("../../test-data/pregenerated_primes.json")
-        ).unwrap();
+pub mod cached {
+    // serialized pregenerated data take so much space, my (virtualized) compiler gets killed
+    // on RAM overuse when trying to `include_str!` the file into the binary. For this reason,
+    // we load and parse on-the-fly the data when it's first accessed.
```
maurges: This comments makes less sense now that they actually fit into ram

maurges (dismissed):

maurges (approved):

PR #175, opened 2026-04-10
[Bugfix]: utils::sqrt no longer maps negative values to 1

### LINKED ISSUE #57: utils::sqrt should probably return an error for negative values
Having it return 1 moves the error into ZK proofs, which might be hard to debug.

PROWLERx15:
Fixes #57 

`utils::sqrt` used to return `1` when `Integer::sqrt_ref()` was `None` (negative input). That hid bad values and made ZK failures harder to trace and debug.

`sqrt` now returns `Option<Integer>` and delegates to `sqrt_ref()`. Call sites in aux key refresh use `.expect()` for our own modulus and treat `None` as a fault when verifying another party’s decommitment.

Tests updated accordingly.

### DIFF
--- a/cggmp24/src/key_refresh/aux_only.rs
+++ b/cggmp24/src/key_refresh/aux_only.rs
@@ -369,7 +369,7 @@ where
         l: L::ELL,
         epsilon: L::EPSILON,
     };
-    let n_sqrt = utils::sqrt(&N).expect("Own modulus N is always positive");
+    let n_sqrt = N.sqrt_ref().ok_or(Bug::NegativeModulus)?;
 
     // message to each party
     for (j, _, d) in decommitments.iter_indexed() {
@@ -456,7 +456,7 @@ where
         &decommitments,
         &shares_msg_b,
         |j, decommitment, proof_msg| {
-            let n_root = match utils::sqrt(&decommitment.N) {
+            let n_root = match decommitment.N.sqrt_ref() {
                 Some(root) => root,
                 None => return true,
             };
--- a/cggmp24/src/key_refresh/aux_only.rs
+++ b/cggmp24/src/key_refresh/aux_only.rs
@@ -369,7 +369,7 @@ where
         l: L::ELL,
         epsilon: L::EPSILON,
     };
-    let n_sqrt = utils::sqrt(&N);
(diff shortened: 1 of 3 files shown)


cggmp24/src/key_refresh/aux_only.rs
```diff
@@ -369,7 +369,7 @@ where
         l: L::ELL,
         epsilon: L::EPSILON,
     };
-    let n_sqrt = utils::sqrt(&N);
+    let n_sqrt = utils::sqrt(&N).expect("Own modulus N is always positive");
```
maurges: We ban the use of unwrap and expect in our code (except in tests). Here you should instead create a new variant of the `key_refresh::Bug` enum and return it

cggmp24/src/utils.rs
```diff
@@ -145,10 +145,10 @@ pub fn iter_peers(i: u16, n: u16) -> impl Iterator<Item = u16> {
...
 
-/// Binary search for rounded down square root. For non-positive numbers returns
-/// one
-pub fn sqrt(x: &Integer) -> Integer {
-    x.sqrt_ref().unwrap_or_else(Integer::one)
+/// Binary search for rounded down square root. Returns `None` for negative
+/// numbers
+pub fn sqrt(x: &Integer) -> Option<Integer> {
+    x.sqrt_ref()
 }
```
maurges: This function is currently just a piece of tech debt - it's a remnant from the time when our big integer backend didn't have the sqrt method. You should remove it completely and use sqrt_ref directly. Obviously the test below is also unnecessary anymore

maurges (changes requested):

cggmp24/src/utils.rs
PROWLERx15: i've removed `utils::sqrt` entirely and replaced it with `sqrt_ref()`. also removed the `test_sqrt` since it is no longer needed

cggmp24/src/key_refresh/aux_only.rs
PROWLERx15: noted. i've replaced it with `Bug::NegativeModulus`

maurges (approved):

maurges:
Looks good, thanks for sending this patch! Now only to wait for the CI to be of the same opinion

PR #176, opened 2026-04-10
fix: validate threshold DKG input parameters

### LINKED ISSUE #39: Threshold DKG doesn't validate input parameters
Threshold DKG doesn't validate that `t <= n`. [`run_threshold_keygen`](https://github.com/dfns-labs/cggmp21/blob/891c30977cc2b107e58844e92da3f5b2adc90c79/cggmp21/src/keygen/threshold.rs#L73C14-L81) should return an error in this case.

Also we should have a test that ensures that setting `t > n` results into expected error.

PROWLERx15:
Fixes #39 

`run_threshold_keygen` did not check `t`, `n`, or `i` up front. Bad
values like `t > n` would run the full protocol and fail later with
unclear errors deep in ZK proofs or share validation.

Added upfront checks at the start of `run_threshold_keygen`:
- `n < 2` → TooFewParties
- `t < 2` → ThresholdTooSmall
- `t > n` → ThresholdTooLarge
- `i >= n` → PartyIndexOutOfRange

All return `KeygenError` via new `Bug` variants immediately, before any
networking or computation.

Tests (all curves):
- `t_greater_than_n` t=5, n=3 → error
- `t_too_small` t=1, n=3 → error
- `n_too_small` t=2, n=1 → error
- `i_out_of_range` i=3, n=3 → error

### DIFF
--- a/tests/tests/it/keygen.rs
+++ b/tests/tests/it/keygen.rs
@@ -138,6 +138,72 @@ where
     validate_keygen_output::<E, Hd>(&mut rng, &key_shares);
 }
 
+cggmp24_tests::test_suite! {
+    test: threshold_keygen_rejects_invalid_params,
+    generics: all_curves,
+    suites: {
+        t_greater_than_n: (5, 3),
+        t_too_small: (1, 3),
+        n_too_small: (2, 1),
+    }
+}
+fn threshold_keygen_rejects_invalid_params<E>(t: u16, n: u16)
+where
+    E: Curve + cggmp24_tests::CurveParams,
+{
+    let mut rng = DevRng::new();
+
+    let eid: [u8; 32] = rng.gen();
+    let eid = ExecutionId::new(&eid);
+
+    let sim_n = n.max(1);
+    let results = round_based::sim::run(sim_n, |i, party| {
+        let party = cggmp24_tests::buffer_outgoing(party);
+        let mut party_rng = rng.fork();
+
+        async move {
+            cggmp24::keygen::<E>(eid, i, n)
+                .set_security_level::<E::SecurityLevel>()
+                .set_digest::<E::Digest>()
+                .set_threshold(t)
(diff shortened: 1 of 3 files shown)


tests/tests/it/keygen.rs
```diff
@@ -138,6 +138,72 @@ where
...
+fn threshold_keygen_rejects_invalid_params<E>(t: u16, n: u16)
+where
+    E: Curve + cggmp24_tests::CurveParams,
+{
+    let mut rng = DevRng::new();
+
+    let eid: [u8; 32] = rng.gen();
+    let eid = ExecutionId::new(&eid);
+
+    let sim_n = n.max(1);
```
maurges, line 159: What's the purpose of this line?

tests/tests/it/keygen.rs
```diff
@@ -138,6 +138,72 @@ where
...
+    let results = round_based::sim::run(sim_n, |i, party| {
```
maurges, line 160: There should be no need for a whole simulator to run one party that is expected to fail before doing any network exchange. Now do we have any examples of this anywhere for you to point you to...

tests/tests/it/keygen.rs
```diff
@@ -138,6 +138,72 @@ where
...
+fn threshold_keygen_rejects_i_out_of_range() {
+    let mut rng = DevRng::new();
+
+    let eid: [u8; 32] = rng.gen();
+    let eid = ExecutionId::new(&eid);
+
+    let (t, n) = (2, 3);
+    let bad_i = n;
+    let result = round_based::sim::run(1, |_sim_i, party| {
+        let party = cggmp24_tests::buffer_outgoing(party);
```
maurges, line 191: Should not be necessary here. Also as above, it would be good to avoid the simulator altogether

maurges (changes requested):

PR #187, opened 2026-04-28
fix: use generic security level in aux precompute instead of hardcoded SecurityLevel128

### LINKED ISSUE #185: secp384r1 signing tests fail due to hardcoded `SecurityLevel128` in aux precompute
**Title:** secp384r1 signing tests fail due to hardcoded `SecurityLevel128` in aux precompute

**Description**

All secp384r1 signing integration tests fail with the following error while all other curves (secp256k1, secp256r1, stark) pass:

```
SigningError(Bug(PiEncElg(psi0, Error(ModPow(BadExponent(ExpSize { exp_size: (~383, 8062), max_exp_size: (2312, 3848) }))))))
```

**Repro**
```
cargo test --workspace
```

**Failing tests**
- `signing::signing_sync::secp384r1::n3`
- `signing::signing_sync::secp384r1::t3n5`
- `signing::signing_with_presigs::secp384r1::t3n5`
- `signing::signing_works::secp384r1::n2`
- `signing::signing_works::secp384r1::n2_reliable`
- `signing::signing_works::secp384r1::n3`
- `signing::signing_works::secp384r1::t2n2`
- `signing::signing_works::secp384r1::t2n3`
- `signing::signing_works::secp384r1::t3n3`

**Root cause**

In `tests/src/lib.rs:246`, the cached aux precompute hardcodes `SecurityLevel128` when computing Pedersen multiexp tables:

```rust
precompute_multiexp_table::<cggmp24::security_level::SecurityLevel128>()
```

However, `CurveParams` for secp384r1 (line 509) uses `SecurityLevel192`. This means secp384r1 signing paths consume aux data precomputed with the wrong security level, causing the `PiEncElg` proof to produce exponents whose size exceeds the expected bounds — hence the `BadExponent(ExpSize)` error.

The slowness observed in secp384r1 tests (repeatedly crossing 60s thresholds) is also a symptom of this mismatch.

**Suggested fix**

Use the generic `L` parameter in `get_aux_inner` for the multiexp precompute instead of hardcoding `SecurityLevel128`.

**Happy to submit a PR if this diagnosis looks correct.**

Mandalorian7773:
Closes #185 

## Problem

`get_aux_inner` in `tests/src/lib.rs` calls `precompute_multiexp_table` with a hardcoded `SecurityLevel128`, ignoring the generic `L` parameter already in scope on the function signature.

This causes all `secp384r1` tests to fail because that curve uses `SecurityLevel192` (via the `CurveParams` trait impl at line 509), creating a type mismatch during aux info precomputation. The bug was silent because CI runs `cargo test --release`, which compiles out all `debug_assert!` calls — so no assertion fired to catch it.

## Fix

**1. One-line bug fix** (`tests/src/lib.rs`):
```diff
-  .precompute_multiexp_table::<cggmp24::security_level::SecurityLevel128>()
+  .precompute_multiexp_table::<L>()
```

**2. Enable debug assertions in release test builds** (`Cargo.toml`):

Added `[profile.release.package.*]` overrides so `debug_assert!` calls inside workspace library code fire during `cargo test --release`, as CI does. Third-party dependencies are unaffected.

```toml
[profile.release.package.cggmp24]
debug-assertions = true

[profile.release.package.cggmp24-keygen]
debug-assertions = true

[profile.release.package.key-share]
debug-assertions = true

[profile.release.package.paillier-zk]
debug-assertions = true
```

This ensures existing `debug_assert!` guards in `paillier-zk`, `key-share`, `cggmp24-keygen`, and `cggmp24/src/signing.rs` will catch invariant violations in future CI runs.

## Impact

- Fixes all `secp384r1` signing tests
- No behavioral change for `secp256k1`, `secp256r1`, or `stark`
- Prevents silent regressions of this class in CI going forward
---

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -54,3 +54,19 @@ num-bigint.opt-level = 3
 num-traits.opt-level = 3
 num-integer.opt-level = 3
 glass_pumpkin.opt-level = 3
+
+# Enable debug assertions for workspace packages in release builds so that
+# debug_assert! calls in library code fire during `cargo test --release`.
+# Note: these overrides only apply within this workspace — binaries that
+# depend on these crates externally are not affected.
+[profile.release.package.cggmp24]
+debug-assertions = true
+
+[profile.release.package.cggmp24-keygen]
+debug-assertions = true
+
+[profile.release.package.key-share]
+debug-assertions = true
+
+[profile.release.package.paillier-zk]
+debug-assertions = true
--- a/tests/src/lib.rs
+++ b/tests/src/lib.rs
@@ -242,9 +242,7 @@ impl PrecomputedKeyShares {
                         crt: None,
                     };
                     params.precompute_crt(&aux.hat_p, &aux.hat_q).unwrap();
-                    params
-                        .precompute_multiexp_table::<cggmp24::security_level::SecurityLevel128>()
-                        .unwrap();
+                    params.precompute_multiexp_table::<L>().unwrap();
                     params
                 })
                 .collect::<Vec<_>>();


Mandalorian7773:
Hi @survived , this PR fixes the secp384r1 failures.
All tests pass locally, would love your review.

Cargo.toml
```diff
@@ -54,3 +54,18 @@ num-bigint.opt-level = 3
 num-traits.opt-level = 3
 num-integer.opt-level = 3
 glass_pumpkin.opt-level = 3
+
+# Enable debug assertions for workspace packages even in release builds, so that
+# debug_assert! calls in library code fire during `cargo test --release` (as used
+# in CI). Third-party dependencies are NOT affected by these overrides.
```
maurges: Nitpick: it's evident that crates not listed here are not affected. I'd rather have a comment like « the binaries using the listed crates are not affected since these options only affect the builds in the context of this workspace », except I phrased it terribly. This fact was not evident to me

maurges:
Looks good, but your patch fails the `cargo fmt` check

Mandalorian7773:
Fixed @maurges

Cargo.toml
```diff
@@ -54,3 +54,19 @@ num-bigint.opt-level = 3
 num-traits.opt-level = 3
 num-integer.opt-level = 3
 glass_pumpkin.opt-level = 3
+
+# Enable debug assertions for workspace packages in release builds so that
+# debug_assert! calls in library code fire during `cargo test --release`.
+# Note: these overrides only apply within this workspace — binaries that
+# depend on these crates externally are not affected.
```
maurges, line 61: Replace with:

> Enable debug assertions in the release builds since we have to run the tests in release mode, because they are too slow

A comment should rarely be an explaination, it should state intent and reason.

And remove the lines between sections, they make them seem unrelated

maurges (changes requested):
Looks good otherwise

PR #194, opened 2026-05-03
signing: identify faulty parties on invalid partial signature

pradhyum6144:
## Problem

When `PartialSignature::combine()` failed, the signing protocol returned
`SigningAborted::SignatureInvalid` with no indication of which party was
responsible. The code had a comment acknowledging this:

> Following the protocol, party should broadcast additional proofs to
> convince others it didn't cheat. However, since identifiable abort is
> not implemented yet, this part of the protocol is missing

## What changed

Each partial signature σ_j must satisfy the linear equation

    σ_j · Γ == m · Δ̃_j + r · S̃_j

against the presignature commitments (public after round 3). Unlike the
`MismatchedDelta` abort — which requires an additional broadcast round to
expose per-party Paillier data — this check needs only the round-4 messages
and the already-public `PresignaturePublicData`. No new protocol round is
required.

The fix iterates over received round-4 messages before calling `combine()`
and collects offending parties into a new `SigningAborted::InvalidPartialSignature`
variant, carrying the standard `Vec<AbortBlame>` with party index and
message IDs as evidence.

`SignatureInvalid` is kept for the residual edge case (r == 0, or σ sum
collapses to zero) where no individual party can be blamed.

## Files changed

- `cggmp24/src/signing.rs`

## Testing

`cargo check --workspace --features all-curves` passes.
`cargo test -p cggmp24 --features all-curves` passes.

### DIFF
--- a/cggmp24/src/signing.rs
+++ b/cggmp24/src/signing.rs
@@ -1463,21 +1463,43 @@ where
     tracer.named_round_begins("Signature reconstruction");
 
     tracer.receive_msgs();
-    let partial_sigs = rounds
+    let round4_msgs = rounds
         .complete(round4)
         .await
         .map_err(IoError::receive_message)?;
     tracer.msgs_received();
 
-    let partial_sigs = partial_sigs
+    // Identify faulty partial signatures using the presignature commitments.
+    // Each σ_j must satisfy σ_j·Γ == m·Δ̃_j + r·S̃_j; any party that fails
+    // this linear check can be pinpointed without an extra broadcast round.
+    tracer.stage("Identify faulty partial signatures");
+    let r = commitments.Gamma.x().to_scalar();
+    if let Some(r) = NonZero::from_scalar(r) {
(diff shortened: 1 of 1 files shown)


pradhyum6144:
Hi @survived, this PR adds identifiable abort for the signing protocol's partial-signature check.

When `combine()` would have returned `None` (invalid aggregate signature), we now iterate over the round-4 messages *before* calling `combine()` and verify each σ_j against the already-public presignature commitments: σ_j·Γ == m·Δ̃_j + r·S̃_j. Any party that fails this linear check is collected into `SigningAborted::InvalidPartialSignature`, which carries the standard `AbortBlame` list with party index and message ID — no extra broadcast round needed.

All existing tests pass locally. Would love your review!

maurges:
Thanks for your interest in this project! Due to the high volume of low-quality contributions we receive, we cannot process all of them and have to close some issues without further expaination.

PR #195, opened 2026-05-03
tests: add secp384r1 external verifier using p384 crate

pradhyum6144:
## What

Replaces the no-op external verifier for `Secp384r1` with a real cross-check against RustCrypto's [`p384`](https://crates.io/crates/p384) crate.

## Why

The `signing_works::secp384r1::*` test suite already calls `E::ExVerifier::verify(...)` after every signing run, but until now it used `Noop` — so the signature bytes were never validated by an independent implementation. Adding a real verifier here closes that gap and gives confidence that cggmp24's P-384 signatures are compatible with the wider RustCrypto ecosystem.

## How

- Added `p384 = { version = "0.13", features = ["ecdsa"] }` to `tests/Cargo.toml`.
- Added `NistP384` struct in `tests/src/external_verifier.rs` implementing `ExternalVerifier<Secp384r1>`:
  - Converts the SEC1-encoded public key to `p384::ecdsa::VerifyingKey`.
  - Writes the cggmp24 signature as 96 bytes (`r ‖ s`) and parses into `p384::ecdsa::Signature`.
  - Calls `VerifyingKey::verify(message, &sig)`, which hashes with SHA-384 internally — matching `Secp384r1`'s `type Digest = sha2::Sha384`.
- Updated `CurveParams for Secp384r1` in `tests/src/lib.rs` to use `NistP384` instead of `Noop`.

All existing tests compile and pass locally (non-secp384r1 suites). The secp384r1 signing tests themselves have a pre-existing failure unrelated to this change (`PiEncElg` max-exponent-size check fails for 384-bit scalars).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -190,6 +190,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "9ba43ea6f343b788c8764558649e08df62f86c6ef251fdaeb1ffd010a9ae50a2"
 
 [[package]]
+name = "base64ct"
+version = "1.8.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "2af50177e190e07a26ab74f8b1efbfe2ef87da2116221318cb1c2e82baf7de06"
+
+[[package]]
 name = "bigdecimal"
 version = "0.3.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -359,6 +365,7 @@ dependencies = [
  "hex",
  "include_dir",
  "lazy_static",
+ "p384",
  "pin-project",
  "rand",
  "rand_dev",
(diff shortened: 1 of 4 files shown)


pradhyum6144:
Hi @survived, this PR adds a real external verifier for secp384r1 using RustCrypto's `p384` crate.

Previously `Secp384r1` used the no-op verifier, so the signature bytes were never validated by an independent implementation. This wires up `NistP384` which cross-checks every signing test output against `p384::ecdsa::VerifyingKey::verify` — the same library the Rust ecosystem uses for P-384.

All tests compile and pass locally. Would love your review!

maurges:
Looks good, but I have to close and reopen it to refresh the stale CI

maurges:
@pradhyum6144 the tests are failing. Not sure if this is caused by the existing code or by passing incorrect parameters to the external library. Possibly this is missing normalization of some sort, if p384 has one

maurges:
Even though the original author has left, this is a serious problem and requires our attention

survived:
It's not that serious, as we do validate all ECDSA signatures using the in-house verifier. External verifier is there just to confirm that signature is also validated by some other library.

survived:
Well, unless there's actually a problem with sigs, I guess we should check it out.

PR #196, opened 2026-05-05
align aff-g names with spec

### LINKED ISSUE #151: CGGMP24 TODOs

AdityaShome:
Summary: fixes #151
Aligns the Paillier affine-operation-in-range proof (aff-g) implementation names with the spec.

Changes:
Renames aff-g SecurityParams fields from l_x and l_y to l and l_prime.
Renames aff-g Data fields from key_j and key_i to n_j and n_i.
Renames aff-g PrivateData fields from nonce and nonce_y to rho and rho_y.
Updates the signing protocol call sites and SecurityParams construction.
Updates the aff-g module docs and doctest example.
Swaps two local equality-check sides in verify so the code order matches the spec text. This does not change behavior because the checks are equality comparisons.

Validation:
cargo fmt --all -- --check
cargo check --workspace --exclude cggmp24-tests
cargo test -p paillier-zk paillier_affine_operation_in_range
cargo test -p paillier-zk --features serde,__internal_doctest --doc paillier_affine_operation_in_range
git diff --check


maurges:

PR #198, opened 2026-05-05
validate parameters before protocol start

AdityaShome:
Summary: related to #151
Adds upfront validation for key generation parameters before entering the DKG protocol runner.

Changes:
Non-threshold keygen now rejects n < 2.
Non-threshold keygen now rejects i >= n.
Threshold keygen now also rejects t < 2.
Threshold keygen now also rejects t > n.
Invalid inputs now fail before networking setup, random sampling, or protocol execution.

Tests:
Added tests that run keygen with invalid parameters using pending incoming messages and drain outgoing messages. The tests pass only if validation happens before the protocol waits on networking.

Validation:
cargo fmt --all -- --check
cargo check --workspace --exclude cggmp24-tests
cargo test -p cggmp24-tests keygen_rejects_invalid_params_before_networking
git diff --check

### DIFF
--- a/cggmp24-keygen/src/lib.rs
+++ b/cggmp24-keygen/src/lib.rs
@@ -218,6 +218,8 @@ where
         R: RngCore + CryptoRng,
         M: Mpc<ProtocolMessage = non_threshold::Msg<E, L, D>>,
     {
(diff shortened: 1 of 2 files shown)


maurges:

PR #203, opened 2026-05-23
fix(paillier-zk): validate nonce is in Z_N* before encrypt_with in verifiers

### LINKED ISSUE #136: Limit size of incoming messages
Currently, the protocol implementation does not limit sizes of incoming messages, which potentially allows attacker to send gigabytish integers that could result into DoS attack. We should consider limiting acceptable ranges from received messages.

ManthanNimodiya:
Closes #136 (partially).

As discussed in #136, the verifiers in `paillier_encryption_in_range`, `paillier_encryption_in_range_with_el_gamal`, and `paillier_affine_operation_in_range` were passing nonce fields (`z2`, `w`, `w_y`) into `encrypt_with` without first checking they belong to the Paillier multiplicative group `Z_N*`. A malicious prover could send an arbitrarily large nonce, triggering expensive Paillier computation before the invalid input is rejected.

This PR adds upfront `in_mult_group_of` checks on those fields before the `encrypt_with` calls, consistent with how other input fields are already validated in these verifiers.

### DIFF
--- a/paillier-zk/src/paillier_affine_operation_in_range.rs
+++ b/paillier-zk/src/paillier_affine_operation_in_range.rs
@@ -406,6 +406,15 @@ pub mod interactive {
             aux.is_in_mult_group(&commitment.t),
         )?;
 
+        fail_if(
+            InvalidProofReason::RangeCheck(17),
+            proof.w.in_mult_group_of(data.key_j.n()),
+        )?;
+        fail_if(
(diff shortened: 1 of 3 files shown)


ManthanNimodiya:
hey @survived, just checking in on my comment above, went ahead and opened a PR for the paillier-zk nonce checks.
can you have a look and let me know for any changes

maurges:

PR #210, opened 2026-05-30
fix(paillier-zk): validate proof bounds before modular exponentiation

### LINKED ISSUE #209: DoS in Π_fac verifier: exponentiation with unbounded proof integers before range checks
Found a DoS issue in the `no_small_factor` (Π_fac) verifier — think it's in the same family as #136.

The problem is in `interactive::verify()`: it does a bunch of `pow_mod`/`combine` calls using proof fields (`z1`, `z2`, `w1`, `w2`, `v`) as exponents *before* it checks whether those values are actually in range. The range checks for `z1`/`z2` only happen at the very end (around line 324), and `w1`, `w2`, `v` have no bounds validation at all.

Since `Integer` has no deserialization size limit, any participant can craft a `fac_proof` inside `MsgRound3` during aux-gen with absurdly large integers. The verifier will happily try to exponentiate with them before rejecting — and modular exponentiation with a multi-thousand-bit exponent is expensive enough to lock up a node.

This is purely an availability issue. No key leakage, no forgery — just DoS.

---

The same pattern appears in two other verifiers:

**`paillier_affine_operation_in_range::interactive::verify`**

https://github.com/LFDT-Lockness/cggmp21/blob/63cedd6b2f75a3aa31faffb8df115d2068db761d/paillier-zk/src/paillier_affine_operation_in_range.rs#L409-L467

`z1`, `z2`, `z3`, `z4`, `w`, `w_y` are all used as exponents or multipliers before the range checks at lines 468–479. `z3`, `z4`, `w`, `w_y` are never range-checked at all.

**`paillier_encryption_in_range_with_el_gamal::interactive::verify`**

https://github.com/LFDT-Lockness/cggmp21/blob/63cedd6b2f75a3aa31faffb8df115d2068db761d/paillier-zk/src/paillier_encryption_in_range_with_el_gamal.rs#L300-L333

`z1` is used in `encrypt_with` and `combine` before its range check at line 335. `z2` and `z3` are not range-checked at all.

---

The fix is straightforward: move the range checks before the expensive math. `paillier_encryption_in_range::interactive::verify` already does this correctly in the same codebase (range checks for `z1`/`z3` happen at lines 273–284, before any exponentiations). Same pattern, just needs to be applied consistently.

PR #203 doesn't cover these paths.

Happy to submit a PR if that's helpful.

gyanranjanpanda:
Fixes #209

The three affected verifiers in `paillier-zk` were running expensive modular exponentiations before they'd checked whether the proof integers used as exponents were even in-bounds. Since `Integer` has no size limit on deserialization, an attacker could craft a proof message with massive integers and pin the verifier with a near-infinite `pow_mod` call — pure DoS, no key material at risk.

The correct pattern is already in the codebase: `paillier_encryption_in_range::interactive::verify` does its range checks upfront before touching any exponentiation. This PR applies the same discipline to the three verifiers that were missing it.

---

### Changes

**`no_small_factor`** (`Π_fac`)

`z1` and `z2` were range-checked at the end of `verify`, after multiple `pow_mod`/`combine` calls that use them. `w1`, `w2`, and `v` had no bounds at all.

- Moved `z1`/`z2` checks before the equality checks
- Added range checks: `w1`, `w2` ∈ ±`2^(l+ε+1) · N̂`, and `v` ∈ ±`n · 2^(l+ε+1) · N̂`

**`paillier_affine_operation_in_range`** (`Π_aff-g`)

Same issue — `z1`/`z2` checked at the end, `z3`/`z4`/`w`/`w_y` unchecked.

- Moved `z1`/`z2` checks before the statement verification block
- Added range checks for `z3`, `z4` (±`N̂ · 2^(l+ε+1)`) and mult-group checks for `w`, `w_y`

**`paillier_encryption_in_range_with_el_gamal`** (`Π_enc-elg`)

`z1` was checked at the very end. `z2` and `z3` had no checks.

- Moved `z1` check before the Paillier operations
- Added mult-group check for `z2` (must be in `Z*_N`) and range check for `z3` (±`N̂ · 2^(l+ε+1)`)

---

### Tests

Each file gets a regression test (`rejects_oversized_response` / `failing_p256_oversized_aux_response`) that injects a crafted `1 << 10_000` integer into the relevant proof field and asserts the verifier rejects it at the correct `RangeCheck` error code before reaching any exponentiation.

All 30 existing tests continue to pass.

### DIFF
--- a/paillier-zk/src/no_small_factor.rs
+++ b/paillier-zk/src/no_small_factor.rs
@@ -289,6 +289,33 @@ pub mod interactive {
                 .map_err(|_| InvalidProofReason::Conversion)?;
             fail_if(InvalidProofReason::RangeCheck(5), n_bits >= 4 * security.l)?;
         }
+
+        let range = (Integer::from(1) << (security.l + security.epsilon)) * data.n_root;
+        fail_if(
+            InvalidProofReason::RangeCheck(12),
+            proof.z1.is_in_half_pm(&range),
+        )?;
+        fail_if(
+            InvalidProofReason::RangeCheck(13),
+            proof.z2.is_in_half_pm(&range),
+        )?;
+
+        let aux_range = (Integer::from(1) << (security.l + security.epsilon + 1)) * &aux.rsa_modulo;
+        fail_if(
+            InvalidProofReason::RangeCheck(14),
+            proof.w1.is_in_half_pm(&aux_range),
+        )?;
+        fail_if(
+            InvalidProofReason::RangeCheck(15),
+            proof.w2.is_in_half_pm(&aux_range),
+        )?;
+
+        let n_at_aux_range = aux_range * data.n;
+        fail_if(
+            InvalidProofReason::RangeCheck(16),
+            proof.v.is_in_half_pm(&n_at_aux_range),
+        )?;
+
         {
             let lhs = aux.combine(&proof.z1, &proof.w1)?;
             let p_to_e = aux.pow_mod(&commitment.p, challenge)?;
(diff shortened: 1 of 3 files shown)


gyanranjanpanda:
@survived  could u review this when u have time

maurges:

PR #229, opened 2026-06-28
Expose set_additive_shift to apply a raw additive key tweak

andresayak:
Adds `SigningBuilder::set_additive_shift(shift: Scalar<E>)`, a public setter for the existing `additive_shift` field that `set_derivation_path` already populates.

## Motivation
Some callers compute the BIP32/SLIP10 non-hardened child tweak `I_L` themselves (e.g. from an externally maintained chain code, when the key share was not generated with an embedded chain code) and need to apply the raw scalar directly. Today the only way to set `additive_shift` is `set_derivation_path`, which re-derives `I_L` internally and requires the share to carry a chain code. This method exposes the same audited shift path for the raw-scalar case.

## What it does
- Feature-gated behind `hd-wallet`, mirroring `set_derivation_path`.
- Sets the same `additive_shift` field; no new crypto. `I_L = 0` is equivalent to no shift.
- The shift is applied unchanged by the existing `signing_t_out_of_n` logic (`X[0] += shift*G`, `x_i += shift` for the first signer), so the signature verifies against `pk + I_L*G`.

Faithful, minimal exposure of an existing code path.

### DIFF
--- a/cggmp24/src/signing.rs
+++ b/cggmp24/src/signing.rs
@@ -535,6 +535,25 @@ where
         Ok(self)
     }
 
+    /// Specifies a raw additive key shift `IL`, so signing is performed under the
+    /// shifted key `sk + IL` and the resulting signature verifies against the
+    /// shifted public key `pk + IL * G`.
+    ///
+    /// This is the same additive-shift mechanism [`set_derivation_path`] uses, but
+    /// the shift scalar is supplied directly instead of being derived from an HD
+    /// path. It lets a caller that computes the BIP32/SLIP10 child tweak itself
+    /// (e.g. from an externally maintained chain code) apply it without the key
+    /// share carrying an embedded chain code. Passing `IL = 0` is equivalent to no
+    /// shift (signs under the original key). The two methods are mutually
+    /// exclusive — calling both keeps the shift set last.
+    ///
+    /// [`set_derivation_path`]: Self::set_derivation_path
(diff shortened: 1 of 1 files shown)


survived:
I can't say that I understand motivation for deriving a shift externally. The only use-case I see is that key was generated externally and [imported to TSS](https://docs.rs/cggmp24/0.7.0-alpha.3/cggmp24/trusted_dealer/index.html). Indeed, [current API](https://docs.rs/cggmp24/0.7.0-alpha.3/cggmp24/trusted_dealer/struct.TrustedDealerBuilder.html) doesn't allow you to set a specific chain code for key shares. If that's the use-case behind your feature-request, then it makes more sense to extend trusted dealer API so it allows specifying the chain code instead generating a random one.

Also, you can always set a chain code in [key shares directly](https://docs.rs/cggmp24/0.7.0-alpha.3/cggmp24/key_share/struct.DirtyKeyInfo.html#structfield.chain_code), the field is part of public API.

survived:
Overall, setting a custom shift feels insecure. If external program that derives it does something malicious, it might compromise security of signing.

Today's API ensures that HD derivation is done properly: chain code is co-generated by all signers, shift is derived by computing a hash function. Outsourcing derivation is dangerous, it removes all these security enforcements, so I don't want it to be part of API, unless there's a strongly motivated use-case that cannot be achieved via current API.

PR #231, opened 2026-07-02
Non-threshold key refresh

Pranavjeet-Naidu:
Adds:  
- 4.5: Key Refresh 
- 4.5.1: non-threshold Key Refresh 

to the spec by splitting the share-refresh half of paper Figure 7 from provisioning
Also includes paper typo fixes and updated presigning preconditions.


survived:
Hi @Pranavjeet-Naidu, thanks for submitting the spec! Could you please set up GPG commit signing? (see [github article](https://docs.github.com/en/authentication/managing-commit-signature-verification)) Otherwise, I will not be able to merge the PR due to the org settings.

spec/main.tex
```diff
@@ -1693,17 +1693,13 @@ \section{Threshold Protocols}
     \item When key generation is requested in an initialized cluster, the signers in that cluster run a distributed key-generation protocol. 
     We have implemented protocols for both $n$-out-of-$n$ key generation (cf.\ Section~\ref{sec:n-of-n-dkg}), as well as $t$-out-of-$n$ key  generation (cf.\ Section~\ref{sec:t-of-n-dkg}).
 
+    \item After a key has been generated, the signers re-randomize their secret key-shares without changing the public key and bind the new shares to a fresh epoch identifier~$\rid$. This key-refresh protocol must be run at least once after key generation before any presignatures are computed, and may be re-run periodically to provide proactive security. It is described in Section~\ref{sec:key-refresh}, we currently provide only the non-threshold (i.e. $n$-out-of-$n$) case.
```
survived: This implies that singers run key refresh protocol right after key generation was finished, however this is not the case. Key refresh protocol can be carried out at any point after keygen was done. It's up to the system requirements and threat model to determine when key refresh is run.

spec/main.tex
```diff
@@ -2000,18 +1996,21 @@ \subsection{Presigning}
...
 
 \begin{description}
     \item[\bf Input.] \
     \begin{itemize}[noitemsep,topsep=0pt]
     \item party index $i$, $0 \le i < n$,
     \item number of signers $n \ge 2$,
-    \item (additive) secret share $x_i \in \Z_q$,
-    \item list of signers' public key shares $\vec X = \{X_j\}_{j\in[n]} \in \E^n$,
+    \item (additive) secret share $x_i \in \Z_q$ (from the most recent key refresh),
+    \item list of signers' public key shares $\vec X = \{X_j\}_{j\in[n]} \in \E^n$ (from the most recent key refresh),
```
survived: normally, key refresh assumes that the old shares are deleted, so it's not necessary to write this down (which suggests that we should preserve all the key shares, that's not the case)

spec/main.tex
```diff
@@ -2000,18 +1996,21 @@ \subsection{Presigning}
...
+The following protocol is based on \cite[Figure~8]{cggmp24}, although we have corrected some typos and eliminated some extraneous parts. (In particular, we do not have identifiable abort.)
+
+\paragraph{Preconditions.} The auxiliary information used below: the parties' Paillier public keys $\vec N$, their Pedersen parameters $\vec R$, and the local Paillier secret key $\sk_i$: is produced once by the provisioning protocol of Section~\ref{section:aux_data_gen} and is unchanged across key refreshes. The (additive) secret share $x_i$, the public key shares $\vec X$, and the refresh epoch identifier $\rid$ are those output by the most recent key-refresh execution (Section~\ref{sec:key-refresh}), which must have been run at least once after key generation. The identifier $\rid$ is incorporated into the session context $\sid$ used below, so that presignatures generated under different refresh epochs are domain-separated; this plays the role of the epoch identifier ($\mathit{epid}$) of~\cite{cggmp24}, in which $\rid$ and the refreshed shares come from key refresh while $\hat N_j, s_j, t_j$ come from provisioning.
```
survived: That's a lot of text to say that "this comes from keygen, and this comes from aux info gen". I'd suggest to simply put items from the list of arguments into "comes from keygen" and "comes from aux info gen" groups.

survived: > The identifier $\rid$ is incorporated into the session context $\sid$

This is not "Preconditions", this explains how the protocol below works. Also, I don't think that the original protocol requires `rid` to be stored nor does it use it in presigning protocol. Am I missing something?

spec/main.tex
```diff
@@ -2240,7 +2239,111 @@ \subsection{Signing}
...
 
 \end{description}
-    
+
+\subsection{Key Refresh}
+\label{sec:key-refresh}
+
+Key refresh re-randomizes the parties' secret key shares without changing the public key, and binds the new shares to a fresh epoch identifier~$\rid$. It corresponds to the share-refresh part of~\cite[Figure~7]{cggmp24}; the generation of Paillier and Pedersen parameters that is interleaved with share refresh in that figure is instead performed once by the provisioning protocol of Section~\ref{section:aux_data_gen}, and is \emph{not} repeated here. Consequently this protocol takes the existing auxiliary information as a precondition and never regenerates $N_i$, $\hat N_i$, $s_i$, or $t_i$.
+
+This protocol must be run at least once after key generation (Section~\ref{sec:n-of-n-dkg}) before any presignatures are computed, and may be re-run periodically to provide proactive security. Each successful execution outputs a new epoch identifier~$\rid$, an updated secret share~$x^*_i$, and updated public key shares~$\vec X^* = (X^*_k)_{k\in[n]}$; the auxiliary information $\hat N_j, s_j, t_j$ (and the local Paillier secret key) is unchanged.
```
survived: There's no such requirement that this protocol needs to be run at least once

spec/main.tex
```diff
@@ -2240,7 +2239,111 @@ \subsection{Signing}
...
+
+\subsubsection{Non-Threshold ($n$-out-of-$n$) Key Refresh}
+\label{sec:n-of-n-refresh}
+
+This protocol is based on the share-refresh part of~\cite[Figure~7]{cggmp24}. We have corrected some typos from that figure.
+The masks used to transmit the new share contributions are derived from pairwise Diffie--Hellman secrets: for each ordered pair $(i, j)$, parties $P_i$ and $P_j$ share the value $y_{i,j} \cdot Y_{j,i} = y_{j,i} \cdot Y_{i,j} = (y_{i,j} y_{j,i}) \cdot G$, from which one-time pads in $\Z_q$ are derived.
```
survived: What masks?

spec/main.tex
```diff
@@ -2240,7 +2239,111 @@ \subsection{Signing}
...
+
+\begin{description}
+    \item[\bf Input.] \
+    \begin{itemize}[noitemsep,topsep=0pt]
+    \item party index $i$, $0 \le i < n$,
+    \item number of signers $n \ge 2$,
+    \item (additive) secret share $x_i \in \Z_q$ from key generation or the previous refresh,
+    \item list of signers' public key shares $\vec X = (X_k)_{k\in[n]} \in \E^n$ from key generation or the previous refresh,
```
survived: would make more sense grouping them in one "Key share:" sub-list

spec/main.tex
```diff
@@ -2240,7 +2239,111 @@ \subsection{Signing}
...
+    \item context separation string $\sid \in \Byte^*$,
+    \item security level $L$ (see Section~\ref{sec:security-params}),
+    \item flag $\var{echo\_enabled} \in \Bit$ indicating whether reliability check is enabled,
+    \item curve $\E$ with generator $G$ of prime order $q$
+    \end{itemize}
+    The auxiliary information $(\hat N_j, s_j, t_j)_{j\in[n]}$ and the local Paillier secret key generated by the provisioning protocol (Section~\ref{section:aux_data_gen}) are unaffected by this protocol, and are therefore omitted from its inputs and outputs.
```
survived: I think it's not necessary to mention this, this was already mentioned in protocol description a few lines above

spec/main.tex
```diff
@@ -2240,7 +2239,111 @@ \subsection{Signing}
...
+
+This protocol must be run at least once after key generation (Section~\ref{sec:n-of-n-dkg}) before any presignatures are computed, and may be re-run periodically to provide proactive security. Each successful execution outputs a new epoch identifier~$\rid$, an updated secret share~$x^*_i$, and updated public key shares~$\vec X^* = (X^*_k)_{k\in[n]}$; the auxiliary information $\hat N_j, s_j, t_j$ (and the local Paillier secret key) is unchanged.
+
+\subsubsection{Non-Threshold ($n$-out-of-$n$) Key Refresh}
+\label{sec:n-of-n-refresh}
+
+This protocol is based on the share-refresh part of~\cite[Figure~7]{cggmp24}. We have corrected some typos from that figure.
```
survived: We have already mentioned once in the spec that we corrected typos from the paper, no need to mention it once again per each protocol

survived: This is the place where you can put a high-level description of the protocol. Keep description functional.

survived: You can mention that this is exactly the protocol from the paper, fig. 7, without aux info (re)generation. All other details seem unnecessary to me.

spec/main.tex
```diff
@@ -2240,7 +2239,111 @@ \subsection{Signing}
...
+    \item[\bf Outputs.] \
+    \begin{itemize}[noitemsep,topsep=0pt]
+    \item refresh epoch identifier $\rid \in \Bit^\kappa$,
+    \item updated (additive) secret share $x^*_i \in \Z_q$,
+    \item updated public key shares $\vec X^* = (X^*_k)_{k\in[n]} \in \E^n$
+    \end{itemize}
+
+    \item[\textbf{Round 1.}] \
+    \begin{algorithmic}[1]
+        \State Sample ephemeral Diffie--Hellman keys $y_{i,0}, \dots, y_{i,n-1} \gets \Z_q$,
```
survived:
```suggestion
        \State Sample ephemeral Diffie--Hellman keys $y_{i,0}, \dots, y_{i,n-1} \gets \Z_q^n$,
```

to match notation in the spec

spec/main.tex
```diff
@@ -2240,7 +2239,111 @@ \subsection{Signing}
...
+               and set $\vec Y_i = (Y_{i,j} = y_{i,j} \cdot G)_{j \in [n]}$
+        \State Sample $x_{i,0}, \dots, x_{i,n-1} \gets \Z_q$ subject to $\sum_{j \in [n]} x_{i,j} = 0$,
```
survived:
```suggestion
        \State Sample $x_{i,0}, \dots, x_{i,n-1} \gets \Z_q^n$ subject to $\sum_{j \in [n]} x_{i,j} = 0$,
```

survived: do not say "sample ... subject to ...", specify exactly what steps we need to do to achieve that property, i.e. 
1. $x_{i,0}, \dots, x_{i, n-2} \gets ℤ_q^{n-1}$
2. and set $x_{i,n-1} = - \sum_{j=0}^{n-2} x_{i,j}$

spec/main.tex
```diff
@@ -2240,7 +2239,111 @@ \subsection{Signing}
...
+               and set $\vec X_i = (X_{i,j} = x_{i,j} \cdot G)_{j \in [n]}$
+        \State For each $j \in [n]$, compute $(A_{i,j}; \tau_{i,j}) \gets \commit{sch}()$;
+               let $\vec A_i = (A_{i,j})_{j \in [n]}$
+        \State Sample $\rid_i, u_i \gets \Bit^\kappa$,
```
survived:
```suggestion
        \State Sample $\rid_i, u_i \gets \Bit^\kappa \times \Bit^\kappa$,
```

spec/main.tex
survived: I see that the original paper does suggest to store `rid` in the key share (indeed looks like "generation" DST), but they never actually make use of it in the signing.

It could be a good idea to inject "generation" separation tag, but I don't want to add `rid` into key share. What we could do is inject a list of public shares into the protocol in place of `rid`.

@maurges @Antoineurban wdyt?

spec/main.tex
```diff
@@ -2240,7 +2239,111 @@ \subsection{Signing}
...
+        \For{$j \in [n] \setminus \{i\}$}
+            \State Assert $(\rid_j, \vec X_j, \vec Y_j, \vec A_j, u_j) \? \in (\Bit^\kappa, \E^n, \E^n, \E^n, \Bit^\kappa)$
+            \State Assert $V_j \? = H(\Encode[hash\_refresh\_com](\sid, j, \vec X_j, \vec Y_j, \vec A_j, \rid_j, u_j))$
+            \State Assert $\sum_{k \in [n]} X_{j,k} \? = \mathcal{O}$
+                   \Comment{\parbox[t]{.45\linewidth}{$\mathcal{O}$ is the identity of $\E$; checks that $P_j$'s contributions sum to zero}}
+        \EndFor
+        \State Compute $\rid = \bigoplus_{j \in [n]} \rid_j$
+        \For{$j \in [n] \setminus \{i\}$}
+            \State Compute the shared secret $W_{i,j} = y_{i,j} \cdot Y_{j,i}$
+            \State Compute $\rho_{i,j} \in \Z_q = H(\Encode[refresh\_mask](\sid, \rid, i, W_{i,j}))$
```
survived: why not simplify?


```suggestion
            \State Compute $\rho_{i,j} \in \Z_q = H(\Encode[refresh\_mask](\sid, \rid, i, y_{i,j} \cdot Y_{j,i}))$
```

survived: Reducing output of hash $\bmod q$ is potentially insecure (depending on the $q$ the result may be biased). You need to use something similar to `ChallengeNI` that we describe in 3.2

spec/main.tex
```diff
@@ -2240,7 +2239,111 @@ \subsection{Signing}
...
+               and for each $j \neq i$ send $C_{j,i}$ to $P_j$
+    \end{algorithmic}
+
+    \item[\textbf{Output.}] \
+    \begin{algorithmic}[1]
+        \State Receive $(\hat\psi_{j,k})_{k \in [n]} \in \Z_q^n$ from all parties,
+               and $C_{i,j} \in \Z_q$ from each $P_j$ with $j \neq i$
+        \For{$j \in [n] \setminus \{i\}$}
+            \State Compute the shared secret $W_{i,j} = y_{i,j} \cdot Y_{j,i}$
+            \State Compute $\rho_{j,i} \in \Z_q = H(\Encode[refresh\_mask](\sid, \rid, j, W_{i,j}))$
```
survived:
```suggestion
            \State Compute $\rho_{j,i} \in \Z_q = H(\Encode[refresh\_mask](\sid, \rid, j, y_{i,j} \cdot Y_{j,i}))$
```

spec/main.tex
```diff
@@ -2240,7 +2239,111 @@ \subsection{Signing}
...
+            \State Compute $x_{j,i} = C_{i,j} - \rho_{j,i} \bmod q$
+            \State Assert $x_{j,i} \cdot G \? = X_{j,i}$ e
```
survived:
```suggestion
            \State Assert $x_{j,i} \cdot G \? = X_{j,i}$
```

Pranavjeet-Naidu:
> Hi @Pranavjeet-Naidu, thanks for submitting the spec! Could you please set up GPG commit signing? (see [github article](https://docs.github.com/en/authentication/managing-commit-signature-verification)) Otherwise, I will not be able to merge the PR due to the org settings.

I've setup GPG signing to be done by default for all commits, once I've addressed all these comments, I'll amend the previous commit and do a force push.

spec/main.tex
maurges: Looking at the paper again: the `rid` value does seem to only be used as domain separation in the zk-proofs inside signing as part of `epid`. The `epid` also includes the list of public shares, and basically every other common public data, which I hope we already hash into the proofs

maurges: It does seem meaningful in the paper, so I think we should add the rid to the share. I'm surprised we didn't before

spec/main.tex
maurges: Yeah, I remember the paper implies this, because they have the aux gen and key refresh in a single protocol, and aux gen needs to run once. For us, with those two protocols separated, we don't need to run key refresh at all

spec/main.tex
```diff
@@ -2240,7 +2239,111 @@ \subsection{Signing}
 \end{inlineAlgorithm}
 
 \end{description}
-    
+
+\subsection{Key Refresh}
+\label{sec:key-refresh}
+
+Key refresh re-randomizes the parties' secret key shares without changing the public key, and binds the new shares to a fresh epoch identifier~$\rid$. It corresponds to the share-refresh part of~\cite[Figure~7]{cggmp24}; the generation of Paillier and Pedersen parameters that is interleaved with share refresh in that figure is instead performed once by the provisioning protocol of Section~\ref{section:aux_data_gen}, and is \emph{not} repeated here. Consequently this protocol takes the existing auxiliary information as a precondition and never regenerates $N_i$, $\hat N_i$, $s_i$, or $t_i$.
```
maurges: I believe in this writeup we describe the key and the aux data separately, so the «existing auxiliary information» is not part of the key share and could be completely omitted from this description

@survived correct me on this because I don't see key share defined at all in the writeup, except as the output of DKG

spec/main.tex
Pranavjeet-Naidu: So, I'll omit this part: 
> This key-refresh protocol must be run at least once after key generation before any presignatures are computed, and may be re-run periodically to provide proactive security. It is described in Section~\ref{sec:key-refresh}, we currently provide only the non-threshold (i.e. $n$-out-of-$n$) case.

and I'll actually also remove the line where it says: 
> we currently provide only the non-threshold (i.e. $n$-out-of-$n$) case.

So, it'll read: 
>Parties may re-randomize their secret key-shares without changing the public key (Section~\ref{sec:key-refresh}); we currently provide only the non-threshold (i.e., $n$-out-of-$n$) case.

spec/main.tex
Pranavjeet-Naidu: Yeah this sentence is irrelevant, I'll remove it completely and let round 3's math handle the explaining.

spec/main.tex
Pranavjeet-Naidu: I'm now adding `rid` in the core key shares, let me know if that's alright.

survived: We used to have rid in the share, we deleted it cuz it was not used

https://github.com/LFDT-Lockness/cggmp21/blob/4c86acd87702db0043d14fa17294f69e890ad1d5/src/key_share.rs#L32-L34

Adding rid to key share would require key share type to be dependent on `L: SecurityLevel`. It's quite an engineering challenge, given that `SecurityLevel` is defined in `cggmp24-keygen`, while key share is defined in `key-share` crate that is independent of the previous one.

Also, `rid` seems a bad design choice as to me. It requires that key refresh protocol implementation (let's say, an external one, developed not by us) updates this field. If the `rid` is not updated, everything works normally, except that domain separation is broken.

Injecting a list of public shares instead of `rid` is way more reliable, because it's not possible to do key refresh without updating list of public shares.

Pranavjeet-Naidu: This does make sense, thank you, I'll defer from having rid in the share, I'll keep the shares as is.

spec/main.tex
maurges: I think this is not necessary to mention anywhere, except maybe once at the top saying how we split the key refresh and aux gen

spec/main.tex
```diff
@@ -2240,7 +2239,111 @@ \subsection{Signing}
...
+    \end{algorithmic}
+
+    \item[\textbf{Round 2.}] \
+    \begin{algorithmic}[1]
+        \State Receive $V_j \in \Bit^\kappa$ from all parties
+        \If{$\var{echo\_enabled}$} \Comment{Reliability check}
+            \State Compute $h_i = H(\Encode[echo](\sid, V_0, \dots, V_{n-1}))$ and send $h_i$ to all parties
+            \State Upon receiving $h_j \in \Bit^\kappa$ from all parties, abort if $h_i \neq h_j$ for some $j \in [n]$
+        \EndIf
+        \State Send $(\rid_i, \vec X_i, \vec Y_i, \vec A_i, u_i)$ to all parties
```
maurges: Slightly weird that the parameters to Send() and H(Encode()) are in a different order

spec/main.tex
Pranavjeet-Naidu: I have made changes to this that should reflect the same once I amend the commit, I should have resolved this comment.

spec/main.tex
Pranavjeet-Naidu: Yes, thanks for pointing that out. Looking at DKG, now I've made it so that: 
rid is present immediately after party index in the hash, and first in the Send() tuple

Pranavjeet-Naidu:
@maurges @survived I've made changes addressing the comments, please take a look. Thank you.

spec/main.tex
survived: Yes, in the spec, we never define explicitly what the "key share" is, but I admit that it could be confusing in the signing to see a long list of input arguments and figure out what comes from where. So it does make sense to group input arguments, for instance.

survived:
@Pranavjeet-Naidu Please, don't do force-pushes unless you absolutely have to. They break the review workflow, we can't see what's changed since the last review.

Pranavjeet-Naidu:
@survived I've added the implementation commits and renamed the branch to reflect both spec and code.

cggmp24/src/key_refresh/share_refresh.rs
survived, line 1: Key refresh should be a separate crate, not a sub-module in the cggmp24. The protocol is like keygen, it can be used in many other places apart from cggmp signing

survived: (the protocol is called key refresh, not share refresh)

cggmp24/src/key_refresh/share_refresh.rs
```diff
@@ -0,0 +1,488 @@
...
+/// Message from round 2
+#[derive(Clone, Serialize, Deserialize, udigest::Digestable)]
+#[serde(bound = "")]
+#[udigest(bound = "")]
+#[udigest(tag = prefixed!("round2"))]
+pub struct MsgRound2<E: Curve, L: SecurityLevel> {
+    /// `rid_i`
+    #[serde(with = "hex")]
+    #[udigest(as_bytes)]
+    pub rid: L::KappaBytes,
```
survived: We usually use HexOrBin for this:

https://github.com/LFDT-Lockness/cggmp21/blob/b870883a1b7a59ff7142071162da5bec3b7f0d3f/cggmp24-keygen/src/non_threshold.rs#L61-L63

cggmp24/src/key_refresh/share_refresh.rs
```diff
@@ -0,0 +1,488 @@
...
+
+/// Message from round 1
+#[derive(Clone, Serialize, Deserialize, udigest::Digestable)]
+#[serde(bound = "")]
+#[udigest(bound = "")]
+#[udigest(tag = prefixed!("round1"))]
+pub struct MsgRound1<D: Digest> {
+    /// $V_i$
+    #[udigest(as_bytes)]
+    pub commitment: digest::Output<D>,
```
survived, line 108: In some other messages, byte slices are serialized with hex (they should use HexOrBin instead), but here and in many other places, they are serialized as is (as list of `u8`). For consistency, we should serialize all slices of bytes in all types of messages in the same way

survived: (I see that there's no consistency in neither keygen nor signing protocols, but we should start by having consistency in new protocols!)

cggmp24/src/key_refresh/share_refresh.rs
```diff
@@ -0,0 +1,488 @@
...
+#[serde(bound = "")]
+pub struct MsgRound3Unicast<E: Curve> {
+    /// $C_{i,j}$
+    pub c: Scalar<E>,
+}
+
+/// Output of the non-threshold share refresh protocol
+pub struct ShareRefreshOutput<E: Curve, L: SecurityLevel> {
+    /// Refreshed core key share
+    pub core: IncompleteKeyShare<E>,
```
survived: I'd call it `share` rather than `core`

cggmp24/src/key_refresh/share_refresh.rs
```diff
@@ -0,0 +1,488 @@
...
+    D: digest::Digest + Clone + 'static,
+    R: RngCore + CryptoRng,
+    M: Mpc<ProtocolMessage = Msg<E, L, D>>,
+{
+    if core.key_info.vss_setup.is_some() {
+        return Err(ShareRefreshReason::NotAdditive.into());
+    }
+    if core.i != i || core.n() != n {
+        return Err(ShareRefreshReason::BadPartyIndex.into());
+    }
```
survived: Is there any point in accepting `i`, `n` as input args if they are always equal to `core.i` and `core.n`?

cggmp24/src/key_refresh/share_refresh.rs
```diff
@@ -0,0 +1,488 @@
...
+    pub rid: L::KappaBytes,
+}
+
+pub async fn run_share_refresh<E, R, M, L, D>(
+    i: u16,
+    n: u16,
+    rng: &mut R,
+    party: M,
+    sid: ExecutionId<'_>,
+    core: &IncompleteKeyShare<E>,
```
survived: I think `share` is a better name for the variable

cggmp24/src/key_refresh/share_refresh.rs
```diff
@@ -0,0 +1,488 @@
...
+    let round3_bc = rounds.add_round(RoundInput::<MsgRound3Broadcast<E>>::broadcast(i, n));
+    let round3_p2p = rounds.add_round(RoundInput::<MsgRound3Unicast<E>>::p2p(i, n));
+    let mut rounds = rounds.listen(incomings);
+
+    tracer.round_begins();
+    let y_secrets: Vec<SecretScalar<E>> = (0..n).map(|_| SecretScalar::random(rng)).collect();
+    let y_points: Vec<Point<E>> = y_secrets
+        .iter()
+        .map(|y| Point::generator() * y)
+        .collect();
```
survived: I'd prefer that the var names match exactly the spec. `y_secrets` should be just `y`, and `y_points` should be `Y` (compiler won't be happy that we use non sneak notation, but you can suppress the warning)

cggmp24/src/key_refresh/share_refresh.rs
```diff
@@ -0,0 +1,488 @@
...
+
+    let mut x_prime: Vec<Scalar<E>> = (0..n - 1).map(|_| Scalar::random(rng)).collect();
+    let sum_head: Scalar<E> = x_prime.iter().sum();
+    x_prime.push(-sum_head);
+    debug_assert!(x_prime.iter().sum::<Scalar<E>>() == Scalar::zero());
```
survived: Usually, clippy is not happy there, you should prefer `debug_assert_eq!`

cggmp24/src/key_refresh/share_refresh.rs
```diff
@@ -0,0 +1,488 @@
...
+
+    let x_prime_points: Vec<Point<E>> = x_prime
```
survived: From naming convention in the spec, this should be called `X_prime`

cggmp24/src/key_refresh/share_refresh.rs
```diff
@@ -0,0 +1,488 @@
...
+        .map(|x| Point::generator() * x)
+        .collect();
+
+    let mut sch_secrets = Vec::with_capacity(usize::from(n));
+    let mut sch_commits = Vec::with_capacity(usize::from(n));
+    for _ in 0..n {
+        let (secret, commit) = schnorr_pok::prover_commits_ephemeral_secret::<E, _>(rng);
+        sch_secrets.push(secret);
+        sch_commits.push(commit);
+    }
```
survived: You can use iterators to do exactly the same thing:

```rust
let (sch_secrets, sch_commits): (Vec<_>, Vec<_>) = 
    std::iter::repeat_with(|| schnorr_pok::prover_commits_ephemeral_secret::<E, _>(rng))
        .take(n.into())
        .unzip();
```

cggmp24/src/key_refresh/share_refresh.rs
```diff
@@ -0,0 +1,488 @@
...
+        let expected = udigest::hash::<D>(&unambiguous::HashRefreshCom {
+            sid,
+            prover: j,
+            decommitment: decom,
+        });
+        let bad_com = com.commitment != expected;
+        let sum: Point<E> = decom.x_prime_points.iter().copied().sum();
+        let bad_sum = sum != Point::zero();
+        bad_len || bad_com || bad_sum
+    });
```
survived, line 340: Is there any reason you're deferring the final verdict until the end? If you have `bad_len`, you can stop verifying early, I see no need to check other criterias.

cggmp24/src/key_refresh/share_refresh.rs
```diff
@@ -0,0 +1,488 @@
...
+    }
+
+    tracer.stage("Calculate challenge rid");
+    let rid = decommitments
+        .iter_including_me(&my_decommitment)
+        .map(|d| &d.rid)
+        .fold(L::KappaBytes::default(), utils::xor_array);
+
+    tracer.stage("Mask refresh shares");
+    let mut masked_out = vec![Scalar::zero(); usize::from(n)];
```
survived: It's really better to use the same names as in the spec, it simplifies the review a lot. Here, it should be `rho`

cggmp24/src/key_refresh/share_refresh.rs
```diff
@@ -0,0 +1,488 @@
...
+        let j_usize = usize::from(j);
+        let dh = &decom_j.y_points[usize::from(i)] * &y_secrets[j_usize];
+        let rho = Scalar::from_hash::<D>(&unambiguous::RefreshMask {
+            sid,
+            rid: rid.as_ref(),
+            party_index: i,
+            dh_shared: &dh,
+        });
+        masked_out[j_usize] = x_prime[j_usize] + rho;
+    }
```
survived: Using index operators inside of loops should be avoided if possible per Rust best-practices. Indexing operators introduce additional boundary check (and panic path) in the compiled code.

survived: I thought that we already had `skip_ith` function in `utils`, but I can't find it. You can add it:

```rust
pub fn skip_ith<I>(i: usize, iter: I) -> impl Iterator<Item = I::Item>
where
    I: IntoIterator
{
    iter.into_iter()
        .enumerate()
        .filter_map(|(j, x)| (i != j).then_some(x))
}
```

then you could rewrite your loop:
```rust
let y_points = decommitments.iter().map(|d| d.y_point[usize::from(i)]);
for (y_point_ji, ((y_secret_j, x_prime_j), masked_out_j))) in y_points.zip(utils::skip_ith(
    i.into(),
    y_secrets.iter().zip(&x_prime).zip(&mut masked_out),
)) {
    let j_usize = usize::from(j);
    let dh = y_point_ji * y_secret_j;
    let rho = Scalar::from_hash::<D>(&unambiguous::RefreshMask {
        sid,
        rid: rid.as_ref(),
        party_index: i,
        dh_shared: &dh,
    });
    *masked_out_j = x_prime_j + rho;
}
```

survived: Or, you could remove the loop altogether:

```rust
let y_points = decommitments.iter().map(|d| d.y_point[usize::from(i)]);
let rhos = y_points
    .zip(utils::skip_ith(i.into(), &y_secrets))
    .map(|(y_point_ji, y_j)| y_point_ji * y_j)
    .map(|dh_shared| Scalar::from_hash::<D>(&unambiguous::RefreshMask {
        sid,
        rid: rid.as_ref(),
        party_index: i,
        dh_shared: &dh_shared,
    });
let masked_out = utils::skip_ith(i.into(), &x_prime)
    .zip(rhos)
    .map(|(x_prime_j, rho_j)| x_prime_j + rho_j)
    .collect::<Vec<_>>();
```

note that in this case, `masked_out` has `n-1` elements, so other code needs to be updated

cggmp24/src/key_refresh/share_refresh.rs
```diff
@@ -0,0 +1,488 @@
...
+    let mut masked_out = vec![Scalar::zero(); usize::from(n)];
+    for (j, _msg_id, decom_j) in decommitments.iter_indexed() {
+        let j_usize = usize::from(j);
+        let dh = &decom_j.y_points[usize::from(i)] * &y_secrets[j_usize];
+        let rho = Scalar::from_hash::<D>(&unambiguous::RefreshMask {
+            sid,
+            rid: rid.as_ref(),
+            party_index: i,
+            dh_shared: &dh,
+        });
```
survived: Reviewing the code, I realized that it might be a good idea to inject both sender and recipient indexes into the hash (currently, we only inject sender index)

cggmp24/src/key_refresh/share_refresh.rs
```diff
@@ -0,0 +1,488 @@
...
+            X: &x_prime_points[k],
+            sch_commit: &sch_commits[k],
+        });
+        let challenge = schnorr_pok::Challenge { nonce: e };
+        sch_proofs.push(schnorr_pok::prove(
+            &sch_secrets[k],
+            &challenge,
+            &x_prime[k],
+        ));
+    }
```
survived: Similarly to suggestions above, you can replace this with an `iter.collect()` which most likely will be cleaner and more functional code.

cggmp24/src/key_refresh/share_refresh.rs
```diff
@@ -0,0 +1,488 @@
...
+
+    tracer.send_msg();
+    outgoings
+        .send(Outgoing::broadcast(Msg::Round3Broadcast(MsgRound3Broadcast {
+            sch_proofs: sch_proofs.clone(),
+        })))
+        .await
+        .map_err(IoError::send_message)?;
+
+    for j in (0..n).filter(|&j| j != i) {
```
survived: use [`utils::iter_peers(i, n)`](https://github.com/LFDT-Lockness/cggmp21/blob/b870883a1b7a59ff7142071162da5bec3b7f0d3f/cggmp24/src/utils.rs#L143-L146)

cggmp24/src/key_refresh/share_refresh.rs
```diff
@@ -0,0 +1,488 @@
...
+        outgoings
+            .send(Outgoing::p2p(
+                j,
+                Msg::Round3Unicast(MsgRound3Unicast {
+                    c: masked_out[usize::from(j)],
```
survived: You can avoid indexing by leveraging iterators, in similar way as was suggested above

cggmp24/src/key_refresh/share_refresh.rs
survived: Actually, it took me a moment to realize that it should be `C`! Different names make spec mapping less trivial.

survived (commented):
Thanks @Pranavjeet-Naidu, that looks solid! I reviewed 3 rounds, left a bunch of stylistic suggestions.

We will need to move the key refresh module to another crate. It would be preferrable to do that in the way that git detects file relocation (so github keeps all suggestions history). It should be possible to do, but if it's challenging, let's finalise the protocol implementation first, and then do the separate crate.

cggmp24/src/key_refresh/share_refresh.rs
Pranavjeet-Naidu: ( that is the reason I didn't bother about consistency but in hindsight, it does seem relevant, will ensure serialisation is consistent ).

cggmp24/src/key_refresh/share_refresh.rs
Pranavjeet-Naidu: apart from using iterators, ill switch the variable names to Tau, A.

cggmp24/src/key_refresh/share_refresh.rs
Pranavjeet-Naidu: Actually yes, no reason to do that. I'll return true immediately on `bad_len`.

cggmp24/src/key_refresh/share_refresh.rs
Pranavjeet-Naidu: No point actually, am changing it.

cggmp24/src/key_refresh/share_refresh.rs
Pranavjeet-Naidu: I forgot to bring it up before, but how do I handle having 2 files named key_refresh: `/src/key_refresh/key_refresh.rs` and `src/key_refresh.rs`?

cggmp24/src/key_refresh/share_refresh.rs
Pranavjeet-Naidu: I'll update the spec as well then? because in the spec, we have: 
`Encode[refresh_mask](sid, rid, i, y_{i,j} · Y_{j,i})`, so only sender i.

Pranavjeet-Naidu:
@survived Regarding the protocol being called key_refresh, the current tests for the protocol are in `tests/tests/it/share_refresh.rs`. Now, if I rename it to `key_refresh.rs`, it conflicts with another file at the same path also called `key_refresh.rs` (this file currently only has the function `aux_gen_works`). Would you be fine if I rename the older `key_refresh.rs` to `aux_gen.rs`?

cggmp24/src/key_refresh/share_refresh.rs
survived: The code should be in a separate crate/folder `cggmp24-key-refresh` (like `cggmp24-keygen`), so there's no name collision within `cggmp24`!

survived: We'll need to clean up `cggmp24` crate to remove all mentions of `key-refresh`, but that's not for this PR

survived:
> Would you be fine if I rename the older `key_refresh.rs` to `aux_gen.rs`?

Absolutely! It should be called this anyway

survived:
I see that you marked many suggestions as "done", but you didn't push any new code, so I can't verify it @Pranavjeet-Naidu

cggmp24/src/key_refresh/share_refresh.rs
survived: Yes, the spec needs to be updated as well!

cggmp24/src/key_refresh/share_refresh.rs
Pranavjeet-Naidu: for now, I'll just just them both stay as key_refresh.rs, and later we can move them around. Thanks!

Pranavjeet-Naidu:
> I see that you marked many suggestions as "done", but you didn't push any new code, so I can't verify it @Pranavjeet-Naidu

I was waiting for the go-ahead on some of the other comments so that I could push all those changes in one commit, I can split them from next time.

spec/main.tex
```diff
@@ -2327,7 +2327,8 @@ \subsubsection{Non-Threshold ($n$-out-of-$n$) Key Refresh}
         \EndFor
         \State Compute $\rid = \bigoplus_{j \in [n]} \rid_j$
         \For{$j \in [n] \setminus \{i\}$}
-            \State Compute $\rho_{i,j} \in \Z_q = \hashtoscalar\bigl(\Encode[refresh\_mask](\sid, \rid, i, y_{i,j} \cdot Y_{j,i})\bigr)$
+            \State Compute $\rho_{i,j} \in \Z_q = \hashtoscalar\bigl(\Encode[refresh\_mask](\sid, \rid, i, j, y_{i,j} \cdot Y_{j,i})\bigr)$
+                   \Comment{\parbox[t]{.45\linewidth}{$i$ is sender, $j$ is recipient; explicit recipient index for domain separation}}
```
survived: Comment is not necessary

cggmp24/src/key_refresh/aux_only.rs
survived, line 1: There should be no changes in `cggmp24/` folder in this PR

tests/tests/it/key_refresh.rs
```diff
@@ -1,43 +1,57 @@
...
-        t2n3: (2, 3, false),
-        t3n5: (3, 5, false),
-        t3n5_reliable: (3, 5, true),
+        n3: (3, false),
+        n5: (5, false),
+        n5_reliable: (5, true),
     }
 }
-fn aux_gen_works<E>(t: u16, n: u16, reliable_broadcast: bool)
+fn key_refresh_works<E>(n: u16, reliable_broadcast: bool)
```
survived, line 16: You should keep the old tst as it was, and add a new test for the key refresh (that does not need aux info generation)

Pranavjeet-Naidu:
@survived @maurges the major changes in the latest commit are: 
- create a separate crate for keyrefresh
- expose both the public api, tests
- work on formatting the comments in specsheet

cggmp24/src/key_refresh/aux_only.rs
survived: @Pranavjeet-Naidu again to reiterate, there should be no changes in `cggmp24/` folder in this PR! Let's focus on the non-threshold key refresh in this PR, then threshold version in the next one, and then figure out how to wire it into the main `cggmp24` library!

cggmp24-key-refresh/src/lib.rs
```diff
@@ -0,0 +1,263 @@
...
+    E: Curve,
+{
+    KeyRefreshBuilder {
+        execution_id: eid,
+        share,
+        reliable_broadcast_enforced: true,
+        tracer: None,
+        _params: core::marker::PhantomData,
+    }
+}
```
survived: You have perfectly reproduced the structure of API that we provide in cggmp24 crate, but I think it will not work here specifically, as eventually we'll also add a threshold version of the protocol.

Threshold key refresh is completely different protocol. Notably, it has different type of the message. The user of the library will have to construct different instances of `round_based::Mpc` for two protocols. So we can't "hide" from the library user that there's actually two protocols. There's no simple mechanism to execute one or another protocol depending on the runtime condition. The protocol being executed is reflected in the type system.

It's quite challenging problem to solve given that we don't have yet the second implementation here, so we can't see the full picture of how the API needs to be organized. What I'd suggest is to just remove this entry point along with constructor for now, and simply expose `cggmp24_key_refresh::non_threshold::run_key_refresh`. It's completely fine, we'll focus on the protocol implementation now, and we'll figure out what's the best API structure later.

survived: Just to give an idea of what the API could look like in the future: I think we'd have two distinct entry function `threshold_key_refresh(eid, key_share) -> Builder` and `non_threshold_key_refresh(eid, key_share) -> Builder`.

We'd need to make sure that `threshold_key_refresh` is only provided a threshold key share, and vice-versa. The best is to make this constraint part of type system. E.g. threshold entry point would accept `ThresholdShare<E>`. That would be a newtype: 

```rust
pub struct ThresholdShare<E>(IncompleteKeyShare<E>);
impl<E> TryFrom<IncompleteKeyShare<E>> for ThresholdShare<E> {
    // ...
}
```

In this way, we introduce a runtime check, and `ThresholdShare` is a witness-type, that guarantees that the check has been performed.

We would do the same witness-type `NonThresholdShare` for the `non_threshold_key_refresh` entry point.

`Builder` then would have to be generic over the key share: `Builder<ThresholdShare>` means that we're preparing to execute a threshold key refresh, and `Builder<NonThresholdShare>` for the non-threshold version.

We would have different `Builder::<T>::run` function with different signature. `Builder::<NonThresholdShare>::run` would accept `round_based::Mpc<ProtocolMsg = non_threshold::msg::Msg>`, and vice-versa. 

I'm not sure if this is the most optimal solution, so I would prefer to have us talk it through on the sync to see if there are easier options for API organisation. For now, let's focus on the protocol!

tests/tests/it/key_refresh.rs
```diff
@@ -48,56 +39,38 @@ where
...
-                aux: aux.into_inner(),
-            }
-            .validate()
-            .unwrap()
-        })
-        .collect::<Vec<_>>();
-
-    // attempt to sign with new shares and verify the signature
+    let original_pk = incomplete_shares[0].shared_public_key;
+    let original_shares: Vec<_> = incomplete_shares
```
maurges: Nitpick: by "shares" we usually mean sharings of a private key, while those are the sharings of the public point. A more fitting name would be "original_public_shares"

tests/tests/it/key_refresh.rs
```diff
@@ -48,56 +39,38 @@ where
...
-    .expect_eq();
+    .into_vec();
 
-    sig.verify(&key_shares[0].core.shared_public_key, &message_to_sign)
-        .expect("signature is not valid");
+    for (i, output) in refreshed.iter().enumerate() {
+        assert_eq!(output.share.shared_public_key, original_pk);
+        assert!(output.share.vss_setup.is_none());
+        assert_ne!(output.share.public_shares, original_shares[i]);
+    }
```
maurges: You could also add the check of `tests::keygen::validate_keygen_output()` (need to make that function public or move it to a common module)

cggmp24-key-refresh/src/non_threshold.rs
```diff
@@ -0,0 +1,501 @@
...
+    E: Curve,
+    L: SecurityLevel,
+    D: digest::Digest + Clone + 'static,
+    R: RngCore + CryptoRng,
+    M: Mpc<ProtocolMessage = Msg<E, L, D>>,
+{
+    if share.key_info.vss_setup.is_some() {
+        return Err(Reason::NotThreshold.into());
+    }
+    let i = share.i;
```
maurges: The `i` in the protocol does not necessarily match the `i` in the share. As an example, the shares can be generated by 6 parties, and then for some reason the 0-th party is forever dropped from the system; then the refresh will be done by parties, whose indexes in the key share go from 1 to 5, but their indexes in the protocol are still 0 to 4.  
Another example: we usually tell that the indexes are derived from the public identities of the parties; well the public identities could well be rotated between invocations.

So the function should still accept the `i` as an argument. For the future threshold version, you will also need a parameter mapping the indexes at keygen. You can look at `signing_t_out_of_n` - there it's called `S`; in other libraries we call it `share_preimages` or `keygen_indexes`

cggmp24-key-refresh/src/non_threshold.rs
```diff
@@ -0,0 +1,501 @@
...
+    let round1 = rounds.add_round(RoundInput::<MsgRound1<D>>::broadcast(i, n));
+    let round1_sync = rounds.add_round(RoundInput::<MsgReliabilityCheck<D>>::broadcast(i, n));
+    let round2 = rounds.add_round(RoundInput::<MsgRound2<E, L>>::broadcast(i, n));
+    let round3_bc = rounds.add_round(RoundInput::<MsgRound3Broadcast<E>>::broadcast(i, n));
+    let round3_p2p = rounds.add_round(RoundInput::<MsgRound3Unicast<E>>::p2p(i, n));
+    let mut rounds = rounds.listen(incomings);
+
+    tracer.round_begins();
+    let y: alloc::vec::Vec<SecretScalar<E>> = (0..n).map(|_| SecretScalar::random(rng)).collect();
+    let Y: alloc::vec::Vec<Point<E>> = y.iter().map(|y_ij| Point::generator() * y_ij).collect();
```
maurges: Nitpick: we prefer the form `collect::<T>` instead of `let x: T = ...collect()` - T is thus understood as a parameter to the collect routine. Also you can add `use alloc::vec::Vec` in the module, like we do in the others

cggmp24-key-refresh/src/non_threshold.rs
```diff
@@ -0,0 +1,501 @@
...
+
+    let X_prime: alloc::vec::Vec<Point<E>> =
+        x_prime.iter().map(|x| Point::generator() * x).collect();
+
+    let (tau, A): (alloc::vec::Vec<_>, alloc::vec::Vec<_>) =
+        core::iter::repeat_with(|| schnorr_pok::prover_commits_ephemeral_secret::<E, _>(rng))
+            .take(usize::from(n))
+            .unzip();
+
+    let mut rid_i = L::KappaBytes::default();
```
maurges, line 236: I wanted to say that the naming is inconsistent in using `_i` or not, but I looked at the other protocols and we were inconsistent there too.. Let's leave as is for now

cggmp24-key-refresh/src/non_threshold.rs
```diff
@@ -0,0 +1,501 @@
...
+    rng.fill_bytes(u_i.as_mut());
+
+    let my_decommitment: MsgRound2<E, L> = MsgRound2 {
+        rid: rid_i,
+        x_prime_points: X_prime.clone(),
+        y_points: Y.clone(),
+        sch_commits: A.clone(),
+        decommit: u_i,
+    };
+    let hash_commit = udigest::hash::<D>(&unambiguous::HashRefreshCom {
```
maurges, line 249: You should add a comment that this corresponds to `V_i` in the writeup. In general we do that when a variable was renamed for some reason (I don't know why we didn't do this in the keygen)

cggmp24-key-refresh/src/non_threshold.rs
```diff
@@ -0,0 +1,501 @@
...
+        commitment: hash_commit,
+    };
+    tracer.send_msg();
+    outgoings
+        .send(Outgoing::broadcast(Msg::Round1(my_commitment.clone())))
+        .await
+        .map_err(IoError::send_message)?;
+    tracer.msg_sent();
+
+    tracer.round_begins();
```
maurges, line 266: It'd be better to add comments saying `// Round 2`, makes it easier to scan the code

cggmp24-key-refresh/src/non_threshold.rs
```diff
@@ -0,0 +1,501 @@
...
+    }
+
+    tracer.stage("Calculate challenge rid");
+    let rid = decommitments
+        .iter_including_me(&my_decommitment)
+        .map(|d| &d.rid)
+        .fold(L::KappaBytes::default(), utils::xor_array);
+
+    let all_decoms: alloc::vec::Vec<_> =
+        decommitments.iter_including_me(&my_decommitment).collect();
```
maurges: This variable is only used in one place, I think it can be inlined there

cggmp24-key-refresh/src/non_threshold.rs
```diff
@@ -0,0 +1,501 @@
...
+            let dh = Y_ji * y_ij;
+            Scalar::from_hash::<D>(&unambiguous::RefreshMask {
+                sid,
+                rid: rid.as_ref(),
+                sender: i,
+                recipient: j,
+                dh_shared: &dh,
+            })
+        });
+    let C: alloc::vec::Vec<Scalar<E>> = utils::skip_ith(usize::from(i), &x_prime)
```
maurges: Now this naming inconsistency is too glaring: they should either be both `rho` and `C`, or both `rhos` and `Cs`

cggmp24-key-refresh/src/non_threshold.rs
```diff
@@ -0,0 +1,501 @@
...
+        .map(|(x_prime_j, rho_j)| x_prime_j + rho_j)
+        .collect();
+
+    tracer.stage("Prove knowledge of x' scalars");
+    let psi_hat: alloc::vec::Vec<_> = X_prime
+        .iter()
+        .zip(&A)
+        .zip(&tau)
+        .zip(&x_prime)
+        .map(|(((X_k, A_k), tau_k), x_k)| {
```
maurges: The `k` notation is not in the writeup, so it could be better as `X_ij_prime, A_ij, tau_ij, x_ij`

cggmp24-key-refresh/src/non_threshold.rs
```diff
@@ -0,0 +1,501 @@
...
+        .filter_map(|((((j, msg_id, msg), Y_ji), X_prime_ji), y_j)| {
+            let dh = Y_ji * y_j;
+            let rho = Scalar::from_hash::<D>(&unambiguous::RefreshMask {
+                sid,
+                rid: rid.as_ref(),
+                sender: j,
+                recipient: i,
+                dh_shared: &dh,
+            });
+            let x_ji = msg.c - rho;
```
maurges: Here I'm having a hard time following the write-up, so it maybe needs more comments. I can deduce that a `_col` is a column of received ij-values with varying `j`; if true, this should still be written out

In the writeup, the `Y` has its subscript swapped compared to here. And `y` could be done with double subscript as well

maurges: Oh and peer_contrib should have a comment that it's `x'_ji in the writeup`

cggmp24-key-refresh/src/non_threshold.rs
```diff
@@ -0,0 +1,501 @@
...
+                proof.verify(A_k, &challenge, X_k).is_err()
+            })
+    });
+    if !blame.is_empty() {
+        return Err(ProtocolAborted::invalid_schnorr_proof(blame).into());
+    }
+
+    tracer.stage("Update key share");
+    let delta: Scalar<E> = peer_contribs.iter().sum::<Scalar<E>>() + x_prime[usize::from(i)];
+    let mut x_star_scalar = *AsRef::<Scalar<E>>::as_ref(&share.x) + delta;
```
maurges: A simple `&share.x + delta` seems to also compile

cggmp24-key-refresh/src/non_threshold.rs
```diff
@@ -0,0 +1,501 @@
...
+    let x_star = NonZero::from_secret_scalar(SecretScalar::new(&mut x_star_scalar))
+        .ok_or(Bug::ZeroSecret)?;
+
+    let mut public_shares = alloc::vec::Vec::with_capacity(usize::from(n));
+    for k in 0..usize::from(n) {
```
maurges: `let public_shares = (0..n).map(|k| ...).collect::<Vec<_>>()`

cggmp24-key-refresh/src/non_threshold.rs
```diff
@@ -0,0 +1,501 @@
...
+        let add: Point<E> = all_decoms.iter().map(|d| d.x_prime_points[k]).sum();
```
maurges: Should probably also be named delta as above. Also, the preferred style is `let delta = ...sum::<Point<E>>();`

cggmp24-key-refresh/src/non_threshold.rs
```diff
@@ -0,0 +1,534 @@
...
+    L: SecurityLevel,
+    D: digest::Digest + Clone + 'static,
+    R: RngCore + CryptoRng,
+    M: Mpc<ProtocolMessage = Msg<E, L, D>>,
+{
+    if share.key_info.vss_setup.is_some() {
+        return Err(Reason::NotThreshold.into());
+    }
+    let n = share.n();
+    debug_assert!(i < n);
```
survived, line 199: We should return a proper error if it's the case

cggmp24-key-refresh/src/non_threshold.rs
```diff
@@ -0,0 +1,534 @@
...
+        .map(|(x_prime_j, rho_j)| x_prime_j + rho_j)
+        .collect::<Vec<_>>();
+
+    tracer.stage("Prove knowledge of x' scalars");
+    let psi_hat = X_prime
+        .iter()
+        .zip(&A)
+        .zip(&tau)
+        .zip(&x_prime)
+        .map(|(((X_ij_prime, A_ij), tau_ij), x_ij)| {
```
survived, line 380: `x_ij` should be `x_ij_prime`

spec/main.tex
```diff
@@ -2240,7 +2254,112 @@ \subsection{Signing}
...
+        \State Compute $\rid = \bigoplus_{j \in [n]} \rid_j$
+        \For{$j \in [n] \setminus \{i\}$}
+            \State Compute $\rho_{i,j} \in \Z_q = \hashtoscalar\bigl(\Encode[refresh\_mask](\sid, \rid, i, j, y_{i,j} \cdot Y_{j,i})\bigr)$
+            \State Compute $C_{j,i} = x'_{i,j} + \rho_{i,j} \bmod q$
+        \EndFor
+        \For{$j \in [n]$}
+            \State Compute $e_{i,j} \in \Z_q = \challengeni{sch}(\sid, i, \rid, X'_{i,j}, A_{i,j})$
+                   and $\hat\psi_{i,j} = \prove{sch}(e_{i,j}; \tau_{i,j}, x'_{i,j})$
+        \EndFor
+        \State Send $(\hat\psi_{i,0}, \dots, \hat\psi_{i,n-1})$,
```
survived, line 2336:
```suggestion
        \State Send $(\hat\psi_{i,0}, \dots, \hat\psi_{i,n-1}) to all parties$,
```

cggmp24-key-refresh/src/non_threshold.rs
```diff
@@ -0,0 +1,534 @@
...
+    for (j, C_j) in utils::iter_peers(i, n).zip(Cs) {
+        outgoings
+            .send(Outgoing::p2p(
+                j,
+                Msg::Round3Unicast(MsgRound3Unicast { c: C_j }),
+            ))
+            .await
+            .map_err(IoError::send_message)?;
+    }
+    outgoings.flush().await.map_err(IoError::send_message)?;
```
survived, line 412: flush is not needed, it is done by `outgoings.send(...).await`

survived: Try to replace with `send_all`

survived: (and make sure there's no more `flush` that is not needed)

cggmp24-key-refresh/src/non_threshold.rs
```diff
@@ -0,0 +1,534 @@
...
+                dh_shared: &dh,
+            });
+            let x_ji_prime = msg.c - rho;
+            if Point::generator() * x_ji_prime != X_prime_ji {
+                masked_blame.push(AbortBlame::new(j, msg_id, msg_id));
+                return None;
+            }
+            Some(x_ji_prime)
+        })
+        .collect::<Vec<_>>();
```
survived, line 462: It's better to split it in two: first compute the `x_ji_prime`, and then in the second part, verify that they match `X_ji_prime` separetely.

cggmp24-key-refresh/src/non_threshold.rs
```diff
@@ -0,0 +1,534 @@
...
+    if !masked_blame.is_empty() {
+        return Err(ProtocolAborted::invalid_masked_share(masked_blame).into());
+    }
+
+    tracer.stage("Validate schnorr proofs");
+    let blame = utils::collect_blame(&decommitments, &sch_proofs_r, |j, decom, msg| {
+        if msg.sch_proofs.len() != usize::from(n)
+            || decom.x_prime_points.len() != usize::from(n)
+            || decom.sch_commits.len() != usize::from(n)
```
survived, line 471: those two checks are already enforced in earlier round

spec/main.tex
```diff
@@ -2240,7 +2254,112 @@ \subsection{Signing}
...
+    \begin{algorithmic}[1]
+        \State Receive $(\hat\psi_{j,k})_{k \in [n]} \in \Z_q^n$ from all parties,
+               and $C_{i,j} \in \Z_q$ from each $P_j$ with $j \neq i$
+        \For{$j \in [n] \setminus \{i\}$}
+            \State Compute $\rho_{j,i} \in \Z_q = \hashtoscalar\bigl(\Encode[refresh\_mask](\sid, \rid, j, i, y_{j,i} \cdot Y_{i,j})\bigr)$
+                   \Comment{\parbox[t]{.3\linewidth}{$j$ is sender, $i$ is recipient}}
+            \State Compute $x'_{j,i} = C_{i,j} - \rho_{j,i} \bmod q$
+            \State Assert $x'_{j,i} \cdot G \? = X'_{j,i}$
+        \EndFor
+        \For{$j \in [n] \setminus \{i\}$ and $k \in [n]$}
```
survived, line 2350: It's better to have two embedded `for` loops for simplicity & to match closely the code

PR #234, opened 2026-07-18
perf(paillier-zk): reduce modular exponentiation cost

### LINKED ISSUE #233: Improve paillier-zk provisioning performance
We profiled CGGMP24 auxiliary-data provisioning with `paillier-zk` 0.7.0-alpha.3 using the num-bigint backend. Approximately 99% of sampled CPU time was modular exponentiation, with 53% in `MultiexpTable::build` and another 17% in the Paillier-Blum prover.

Two backend-generic changes produced measurable improvements:

- Build radix-256 multiexponentiation powers incrementally instead of recomputing `base^(256^i) mod N`: provisioning fell from 64.9s to 31.8s serially.
- Use `fast_paillier::utils::CrtExp` in `paillier_blum_modulus::interactive::prove`, where the prover already knows `p` and `q`: serial provisioning fell further to 29.1s.

All 27 paillier-zk library tests pass, the Rug/GMP backend compiles, and an end-to-end three-party provisioning test passes with both changes. I plan to submit these as two signed and DCO-signed-off commits.

noah-dfns:
Closes #233.

Optimize aux info generation:
- Propagate CRT and multiexp optimization to all the proofs
- Improve efficiency of multiexp table construction

### DIFF
--- a/cggmp24/src/key_refresh/aux_only.rs
+++ b/cggmp24/src/key_refresh/aux_only.rs
@@ -13,8 +13,7 @@ use crate::{
     key_share::{AuxInfo, DirtyAuxInfo, PedersenParams, Validate},
     progress::Tracer,
     security_level::SecurityLevel,
-    utils,
-    utils::{collect_blame, AbortBlame},
+    utils::{self, collect_blame, AbortBlame},
     zk::ring_pedersen_parameters as π_prm,
     ExecutionId,
 };
@@ -187,19 +186,17 @@ where
 
     tracer.stage("Build Paillier key");
     let N = &p * &q;
+    let N_crt = paillier_zk::fast_paillier::utils::CrtExp::build_n(&p, &q).ok_or(Bug::BuildNCrt)?;
 
     tracer.stage("Build Pedersen params");
-    let (pedersen_params, phi_hat_N, lambda) = utils::generate_pedersen_params(rng, hat_p, hat_q)?;
+    let (my_pedersen_params, phi_hat_N, lambda) =
+        utils::generate_pedersen_params(rng, hat_p, hat_q)?;
 
     tracer.stage("Prove Πprm (ψˆ_i)");
     let hat_psi = π_prm::prove::<{ crate::security_level::M }, D>(
         &unambiguous::ProofPrm { sid, prover: i },
         &mut rng,
-        π_prm::Data {
-            N: &pedersen_params.hat_N,
-            s: &pedersen_params.s,
-            t: &pedersen_params.t,
-        },
+        &my_pedersen_params,
         &phi_hat_N,
         &lambda,
     )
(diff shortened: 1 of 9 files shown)


maurges (dismissed):

paillier-zk/src/multiexp.rs
```diff
@@ -40,14 +40,20 @@ impl MultiexpTable {
...
+        let radix = Integer::from(256u32);
+        let mut s_power = s.pow_mod_ref(&Integer::one(), &N)?;
         for i in 0..k_x {
-            let B_to_i = Integer::u_pow_u(B, i);
-            s_table.push(s.pow_mod_ref(&B_to_i, &N)?);
+            s_table.push(s_power.clone());
+            if i + 1 < k_x {
+                s_power = s_power.pow_mod_ref(&radix, &N)?;
+            }
         }
```
survived: I think this code became a bit too complicated. I'll give it a shot to simplify it.

maurges: I was thinking to nitpick variable names, and in the end you didn't even change it.. I find the code fine, but I don't mind your change too

maurges (dismissed):

cggmp24/src/key_refresh/aux_only.rs
```diff
@@ -320,29 +317,59 @@ where
...
+        .iter()
+        .map(|d| PedersenParams {
+            hat_N: d.hat_N.clone(),
+            s: d.s.clone(),
+            t: d.t.clone(),
+            multiexp: None,
+            crt: None,
+        })
+        .collect::<Vec<_>>();
+    pedersen_params.insert(i.into(), my_pedersen_params);
```
maurges, line 344: I feel like the insert could be done on the iterator instead of the vector, but I didn't find a relevant method and I don't want to force you to write a new utility function for a marginal gain  
Edit: disregard that at all, since it was copied from below

maurges: \+ I liked the name parties_pedersen more

cggmp24/src/key_refresh/aux_only.rs
```diff
@@ -320,29 +317,59 @@ where
...
+            .map(|d| d.N.clone())
+            .collect::<Vec<_>>(),
+        pedersen_params,
+        security_level: std::marker::PhantomData,
+    };
+    if compute_multiexp_table {
+        tracer.stage("Precompute multiexp tables");
+        aux.precompute_multiexp_tables()
+            .map_err(Bug::BuildMultiexpTables)?;
+    }
```
maurges, line 359: It seems this could be moved to after the check below has run

maurges (dismissed):
Good change at large, left two nitpicks

cggmp24/src/key_refresh/aux_only.rs
survived: I explicitly divided that check into two, because the second part uses multiexp&crt internally, so it benefits from them being generated

cggmp24/src/key_refresh/aux_only.rs
survived: `pedersen_params` matches the name of field in `DirtyAuxInfo`

cggmp24/src/key_refresh/aux_only.rs
maurges: Oh yes, I missed the usage

maurges (approved):

PR #235, opened 2026-07-19
perf(paillier-zk): retain Montgomery multiexp context

noah-dfns:
## Summary
- cache multiexponentiation table values in a reusable Montgomery domain
- retain the existing implementation as an exact fallback for even moduli
- keep serialized fields unchanged and lazily rebuild the cache after deserialization
- pin the permissively licensed fast-paillier backend from LFDT-Lockness/fast-paillier#42

This is stacked on #234. Review the final commit; after #234 merges, this branch can be rebased to leave only the cache change.

## Performance
The cache alone improved isolated signing tests by about 4%. On top of the Dashu backend, full-suite user CPU fell from 223.17s to a 210.10s median, another 5.9%. The production 512/448-bit table adds approximately 48 KiB of runtime cache.

## Validation
- 28 paillier-zk unit tests
- 6 doctests
- serialization round trip and lazy cache rebuild
- odd-modulus optimized path and even-modulus fallback
- Clippy with warnings denied
- downstream 130-test integration suite, twice

Signed-off-by: Noah <noah@dfns.co>

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -178,6 +178,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "4c7f02d4ea65f2c1853089ffd8d2787bdbc63de2f0d29dedbcf8ccdfa0ccd4cf"
 
 [[package]]
+name = "base16ct"
+version = "1.0.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "fd307490d624467aa6f74b0eabb77633d1f758a7b25f12bceb0b22e08d9726f6"
+
+[[package]]
 name = "base64"
 version = "0.13.1"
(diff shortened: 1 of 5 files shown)


maurges:
I think the changes are too complicated for the improvement they bring. Running the tests on my machine, I get **20**mins __no__ change vs **30**mins __with__ the change. The test time is still dominated by prime generation, which has a very wide distribution which can swing time measurements significantly. For one example, while one change included here (that to fast-paillier) makes the fast-paillier benchmarks 1.5 times faster, it has no noticable effect on the runtime of these tests

We would still like the low-hanging-fruit improvements to the performance; Denis has been sniped into working on such improvements inspired by the other PR, passing the CRT tables to proofs. It's going to shave off a good amount of time off the protocol, but also won't affect the test performance significantly

